PR #1, opened 2019-01-24
Fix build for android devices

dkanada:
I haven't figured out how to build this as a normal java library but this should get it working for android devices. Once this gets merged I can use the library in the android and androidtv repositories, allowing us to publish on F-Droid.


dkanada:
correction: androidtv still has a few libraries that will block an F-Droid release for now

dkanada:
I think F-Droid supports it for when a user sideloads the F-Droid apk onto their television, so it isn't a very large userbase. We could probably just ignore AndroidTV for F-Droid.

PR #3, opened 2019-02-22
Build project with cordova compatibility

dkanada:


JustAMan:
Why downgrade lots of things? Please explain.

dkanada:
The version of cordova in my PR only supports v27 and lots ofbuild errors showed up when the apiclient was on v28 instead.

PR #7, opened 2019-03-08
Remove registration check for features

thornbill:
:unlock: Unlock all the things!


android/src/main/java/mediabrowser/apiinteraction/connectionmanager/ConnectionManager.java
```diff
@@ -751,39 +751,11 @@ public void ExchangePin(PinCreationResult pin, final Response<PinExchangeResult>
     }
 
     public void GetRegistrationInfo(final String featureName, String serverId, String localUsername, final Response<RegistrationInfo> response) {
```
cvium: Shouldn't it just be removed? The API method has already been nulled on the server side

thornbill: Yes, but that could introduce errors if the apps are using this method. I have added an annotation to indicate the method is deprecated.

PR #9, opened 2019-03-24
Add Android implementation of ILogger

thornbill:
Add an implementation of `ILogger` that uses the Android Log utility. This is to replace the usage of `ConsoleLogger` in the Android TV app which uses `System.out` and therefore is not properly tagged in Android logs.

### DIFF
--- a/android/src/main/java/mediabrowser/apiinteraction/android/AndroidLogger.java
+++ b/android/src/main/java/mediabrowser/apiinteraction/android/AndroidLogger.java
@@ -1,0 +1,51 @@
+package mediabrowser.apiinteraction.android;
+
+import android.util.Log;
+
+import mediabrowser.model.logging.ILogger;
+
+/**
+ * ILogger implementation using the Android Log utility.
+ */
+public class AndroidLogger implements ILogger {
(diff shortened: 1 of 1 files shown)


android/src/main/java/mediabrowser/apiinteraction/android/AndroidLogger.java
```diff
@@ -0,0 +1,51 @@
+package mediabrowser.apiinteraction.android;
```
JustAMan: can we start transitioning to `jellyfin` package?

dkanada: That should probably be done all at once here or it would be very weird. Some packages would be in one folder and others in a different folder.

thornbill: I would like to do that all at once. It really bugs me that they didn’t follow standard namespace conventions i.e. `org.jellyfin.package`. At the same time I plan to divide the project into two gradle modules so we can have a proper Java client without the Android dependencies.

PR #10, opened 2019-03-24
Update autodiscovery message

thornbill:
Use "who is JellyfinServer?" instead of "who is EmbyServer?" for autodiscovery.


cvium:
Fyi, the server supports both

JustAMan:
> Fyi, the server supports both

But this was intended for compatibility with older apps, so updating it in new apps makes sense, right? 😉

PR #11, opened 2019-03-24
Add device name from settings

thornbill:
* Add support for pulling the user set device name from `Settings.Global` in favor of the device's model name.
* Add check for null `Cursor` and properly close it after use.

refs: https://stackoverflow.com/a/44460486/1373956


JustAMan:
@thornbill can you use `Settings.Global.DEVICE_NAME` instead of hardcoding `"device_name"` (per same SO answer)?

thornbill:
No. I tried that but the constant is unavailable in older SDK versions. Android Studio yelled at me.

PR #12, opened 2019-04-04
Refactor project structure

thornbill:
**Anything currently using the apiclient will need updated to reflect these changes.**

* Moves namespace to `org.jellyfin.apiclient`
* Renames `apiinteraction` package to `interaction`
* Splits Android specific code into separate gradle module
* Moves classes in `android` package to `interaction`
* Moves `AndroidLogger` to `logging` package


thornbill:
There are now corresponding updates in jellyfin-android and jellyfin-androidtv so this is ready for review.

library/src/main/java/org/jellyfin/apiclient/interaction/BaseApiClient.java
```diff
@@ -279,7 +277,7 @@ public final String GetApiUrl(String handler)
 		return GetApiUrl(handler, new QueryStringDictionary());
 	}
 
-    protected mediabrowser.apiinteraction.http.HttpHeaders HttpHeaders = new HttpHeaders();
+    protected org.jellyfin.apiclient.interaction.http.HttpHeaders HttpHeaders = new HttpHeaders();
```
dkanada: There are a couple declarations with the full name listed out that look like they can be shortened.

PR #15, opened 2019-07-13
Remove dependency on dcendents

dkanada:
@thornbill I have no issues building the client without dcendents, is that plugin required for a specific reason?


thornbill:
It’s needed for jitpack to build the Android package https://www.jitpack.io/docs/ANDROID/

PR #16, opened 2019-07-13
Remove some old code artifacts and two insecure hash methods

dkanada:
I also couldn't build without changing the import statement for the LruCache class.


dkanada:
I still think using sha256 or something as a replacement isn't a terrible idea but these two methods are outdated regardless.

library/src/main/java/org/jellyfin/apiclient/interaction/websocket/ApiWebSocket.java
```diff
@@ -290,63 +287,30 @@ private void OnGeneralCommand(String json)
...
-                tangible.RefObject<Long> tempRef_expected = new tangible.RefObject<Long>(expected);
-                LongHelper.TryParseCultureInvariant(timeoutMs, tempRef_expected);
-                expected = tempRef_expected.argValue;
-
                 MessageCommand command = new MessageCommand();
-
                 command.setHeader(header);
                 command.setText(text);
-                command.setTimeoutMs(expected);
+                command.setTimeoutMs(Long.parseLong(timeoutMs));
```
thornbill: We should probably catch `NumberFormatException`s here like the `TryParseCultureInvariant` method did.

library/src/main/java/org/jellyfin/apiclient/model/dlna/DeviceProfile.java
```diff
@@ -473,16 +473,15 @@ public DeviceProfile()
 
 	public final TranscodingProfile GetAudioTranscodingProfile(String container, String audioCodec)
 	{
-		container = StringHelper.TrimStart((container != null) ? container : "", '.');
-
+		container = (container != null) ? container.replace(".", "") : "";
```
thornbill: This no longer removes leading whitespace. May introduce new issues?

library/src/main/java/org/jellyfin/apiclient/interaction/connectionmanager/ConnectionManager.java
```diff
@@ -690,7 +689,7 @@ String NormalizeAddress(String address) throws IllegalArgumentException {
             throw new IllegalArgumentException("address");
         }
 
-        if (StringHelper.IndexOfIgnoreCase(address, "http") == -1)
+        if (!address.contains("http"))
```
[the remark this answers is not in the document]
dkanada: I used patterns for all these cases but I don't plan on actually fixing the root issue in this pull request - might fix it eventually.

library/src/main/java/org/jellyfin/apiclient/interaction/websocket/ApiWebSocket.java
dkanada: I think the best short term plan is to pass the exception up a level so the app can handle the error with its chosen logging implementation.

library/src/main/java/org/jellyfin/apiclient/model/dlna/DeviceProfile.java
dkanada: I honestly don't think this one will be an issue, do you think so?

PR #20, opened 2019-08-14
Remove wake on lan classes

dkanada:


library/src/main/java/org/jellyfin/apiclient/model/apiclient/ServerInfo.java
```diff
@@ -97,7 +88,6 @@ public final void setLastConnectionMode(ConnectionMode value)
 
 	public ServerInfo()
```
thornbill: This constructor is no longer needed.

PR #23, opened 2019-09-04
Remove wan address and connection mode

dkanada:


joshuaboniface:
@thornbill We good with this one now?

thornbill:
Nope. Without corresponding changes in the Android TV client, this will cause build failures there. Although this is **not** required for the WAN removal server side.

joshuaboniface:
Roger, removing from the 10.4.0 project for now.

PR #24, opened 2019-09-11
Set up CI with Azure Pipelines

thornbill:


EraYaN:
Did you test it? It seems to me that it might be missing a `-job` key.

thornbill:
Azure Pipelines are new to me so I definitely could have missed something, but it seems to be working. I tested on my fork here: https://github.com/thornbill/jellyfin-apiclient-java/pull/1

EraYaN:
All right ship it!

PR #26, opened 2019-09-19
Use ApiClient methods for url generation

thornbill:
This addresses the issue in https://github.com/jellyfin/jellyfin-androidtv/issues/155 where the initial user lookup uses the old "mediabrowser" api endpoint.

### DIFF
--- a/library/src/main/java/org/jellyfin/apiclient/interaction/connectionmanager/ValidateAuthenticationResponse.java
+++ b/library/src/main/java/org/jellyfin/apiclient/interaction/connectionmanager/ValidateAuthenticationResponse.java
@@ -1,5 +1,6 @@
 package org.jellyfin.apiclient.interaction.connectionmanager;
 
+import org.jellyfin.apiclient.interaction.ApiClient;
 import org.jellyfin.apiclient.interaction.EmptyResponse;
 import org.jellyfin.apiclient.interaction.Response;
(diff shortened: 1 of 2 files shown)


library/src/main/java/org/jellyfin/apiclient/interaction/connectionmanager/ValidateAuthenticationResponse.java
```diff
@@ -36,14 +33,19 @@ public void onResponse(String jsonResponse) {
...
+            if (client != null) {
+                String apiUrl = client.GetApiUrl("Users/" + server.getUserId());
+                apiUrl = client.AddDataFormat(apiUrl);
+                request.setUrl(apiUrl);
+
+                httpClient.Send(request, new ValidateAuthenticationInnerResponse(connectionManager, jsonSerializer, server, response));
+                return;
+            }
         }
+        response.onResponse();
```
dkanada: Shouldn't this stay in an `else` block? The inner response may not have been triggered yet.

thornbill: The `return` above should be used in the event that this should not be called. Otherwise we would need to handle the case where `client` is `null` also. I thought this approach would be cleaner.

I tested this in Android TV and it seems to be working as expected there for what it's worth.

PR #30, opened 2019-11-29
Introduce Enum for Types and deprecate strings

AndreasGB:
This deprecates the BaseItemDto.getType() methods and introduces a new EBaseItemType Enum, which gives this field an explicit value range instead of the old method of having strings scattered across the library.

Please release this as v0.5.0, as the AndroidTV App refactoring (jellyfin/jellyfin-androidtv#217) of this will use this as the version it references, thanks :)


android/build.gradle
```diff
@@ -9,7 +9,7 @@ android {
         minSdkVersion 19
         targetSdkVersion 28
         versionCode 26
-        versionName "0.4.0"
+        versionName "0.5.0"
```
dkanada: I would wait to bump the version until we actually want to release a new version.

library/src/main/java/org/jellyfin/apiclient/model/dto/BaseItemDto.java
```diff
@@ -1026,15 +1026,27 @@ public final void setParentId(String value)
...
 	}
 
+	public final void setEBaseItemType(EBaseItemType Type) {
+		this.Type = Type.name();
+	}
+
+	public final EBaseItemType getEBaseItemType() {
+		return EBaseItemType.valueOf(Type);
+	}
+
```
dkanada: Not a fan of the `E` prefix for this variable, @thornbill what do you think?

android/build.gradle
AndreasGB: Hmm, but it's a new version that then needs to be referenced by the Android TV build.

library/src/main/java/org/jellyfin/apiclient/model/dto/BaseItemDto.java
thornbill: I agree. I’m not a fan of using prefixes to indicate an object type.

AndreasGB: Okay, changed :)

android/build.gradle
dkanada: I hadn't seen that, works for me. We should only bump to 0.4.1 though, especially since 1.0.0 should only happen when we can generate the client with OpenAPI.

AndreasGB: I would still recommend 0.5.0, since the smallest digit is for bugfixes only and there is no rule that you have to roll over from 0.9.0 to 1.0.0. As I understand it, we can keep incrementing the second number as far as we like. 

Otoh, everything with leading zero is by definition unstable anyway, so we might also go with your recommendation. 

@thornbill Thoughts on this?

dkanada: I'm not personally a fan of numbers past nine for minor versions but I suppost the Android TV app is already using them so it's irrelevant here.

AndreasGB: I get that it doesn't look that nice aesthetically, but I think it is the correct thing to do according to https://semver.org/

PR #33, opened 2020-02-03
Refactor PersonType into a proper enum

AndreasGB:


AndreasGB:
Wait, shit, I used the wrong identity again...

AndreasGB:
Okay, should now be the correct one

library/src/main/java/org/jellyfin/apiclient/model/entities/PersonType.java
```diff
@@ -3,38 +3,7 @@
...
-    public static final String Producer = "Producer";
-    /**
-     The conductor
-     */
-    public static final String Conductor = "Conductor";
-    /**
-     The lyricist
-     */
-    public static final String Lyricist = "Lyricist";
+Actor, Director, Composer, Writer, GuestStar, Producer, Conductor, Lyricist, Other
```
nielsvanvelzen: Formatting looks a bit off. I think a tab (or 4 spaces) is missing at the start of the row.
If you use IntelliJ or Android Studio just use the built-in formatter

library/src/main/java/org/jellyfin/apiclient/model/dto/BaseItemPerson.java
```diff
@@ -58,15 +60,31 @@ public final void setRole(String value)
 	 <value>The type.</value>
 	*/
 	private String Type;
+	@Deprecated
 	public final String getType()
```
nielsvanvelzen: Can you remove the deprecated methods or are they still used somewhere?

library/src/main/java/org/jellyfin/apiclient/model/entities/PersonType.java
AndreasGB: Whoops, fixing. I also deleted the nondescriptive comment at the file start now as well.

library/src/main/java/org/jellyfin/apiclient/model/dto/BaseItemPerson.java
nielsvanvelzen: We decided to leave them in because the Android (not TV) app uses the master branch and we don't want to cause issues over there.

It shouldn't actually cause issues as the Android app doesn't use these models but just to be sure :)

AndreasGB: Unclear, Android TV doesn't, other clients still might.

PR #35, opened 2020-02-07
Added checks that hopefully catch the dumbest misusages of logger

AndreasGB:


library/src/main/java/org/jellyfin/apiclient/logging/ConsoleLogger.java
```diff
@@ -5,39 +5,39 @@
...
     }
 
     @Override
-    public void ErrorException(String message, Exception exception, Object... paramList) {
-        LogException(message, exception, paramList);
+    public void ErrorException(String formatString, Exception exception, Object... paramList) {
+        LogException(formatString, exception, paramList);
     }
 
     private void LogException(String message, Exception exception, Object... paramList) {
```
nielsvanvelzen: You forgot this one 👀

AndreasGB: Actually didn't touch that file at all besides automatic rename, as it isn't used anywhere, but will now

PR #37, opened 2020-02-09
Try both https and http when no protocol is set

nielsvanvelzen:
Changes:
- Formatted the files I worked on (includes import cleaning)
- Try multiple server urls in the connect function
- The `NormalizeAddress` now trims the input
- The `NormalizeAddress` now properly detects if a protocol is set
- The `NormalizeAddress` can return multiple urls (Could add a default port too in the future)
- The `ServerLocator` now doesn't normalize because the connect function does that already..

Tested both manual input and server detection to work properly.


Artiume:
Thanks for the PR, this is the first bug I ever encountered with AndroidTV since I never append the protocol haha

thornbill:
I would really like for there to be unit tests to cover the address normalization. Do you think that would be possible to add here @nielsvanvelzen?

nielsvanvelzen:
The `ConnectionManager` class takes quite some parameters to initialize but it should be possible. Will take a look at this soon.

PR #39, opened 2020-02-17
Update Java-WebSocket to 1.4.0 and some misc fixes

nielsvanvelzen:
Looking at the logs the app showed me it tried to connect but the connection would close within a few sec. I noticed the path was set to `/embysocket` but web connects to `/socket` so I changed that.
It still didn't work... What could happen? Hmm looking at the code I noticed the `onClose` method gets 3 parameters: `int i, String s, boolean b` but they aren't logged.. Let's log them!

The error was in string s and said: `draft org.java_websocket.drafts.Draft_10@236b291 refuses handshake`.. Well.. What's draft 10? It's an old version of the websocket spec! Can we upgrade it? Our library supports setting the draft as parameter in the constructor so I set it to the latest version available, this was 76.. Still didn't work :(
Looking at Wikipedia I saw there's a newer version for the protocol so let's look at the library version... We use 1.3.0 and the latest version is 1.4.0. I tried upgrading but breaking changes, hmm lets try a slightly older version first for testing. I changed it to 1.3.4 as it introduced Draft_6455 and deprecated all the other ones..

I compiled this and.. well... it worked!

I proceeded to upgrade to the latest version and added an additional fix when determining the websocket url.

Example how it would look in the webclient when this fix is applied in Android TV:
image

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'java'
 apply plugin: 'maven'
 
 dependencies {
-    implementation 'org.java-websocket:Java-WebSocket:1.3.0'
+    implementation 'org.java-websocket:Java-WebSocket:1.4.0'
     implementation 'com.google.code.gson:gson:2.8.2'
     testImplementation 'junit:junit:4.12'
 }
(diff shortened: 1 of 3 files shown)


Artiume:
Could this be impacted by setting a base url?

nielsvanvelzen:
It uses the URL already available and replaces the `http` part with `ws` (https becomes wss). Baseurl should work because it doesn't remove anything from the path.

I can not confirm that though.

PR #45, opened 2020-04-30
Remove image loading utilities

nielsvanvelzen:
Caching images is **not** a function that we should handle ourselves. Clients should instead use a library like Glide to handle this (as they can do it much better).

Looking at the source code of both ATV and Android only 1 instance was found using the `getBitmap` function in the Android app.


nielsvanvelzen:
The codeclimate issue is about a method that is used by clients and was not introduced in this PR. Please ignore it.

PR #49, opened 2020-04-30
Add "model" package to it's own module

nielsvanvelzen:
This PR adds the model package to it's own module. This way third-party libraries/applications can use the models with their own HTTP client and things like that.

Additional changes made:
- The extra serialization subpackage was kept in "library" (but moved)
- Removed the use of tangible
- library now uses the "java-library" plugin so it can use the "api" dependency type
- Added the `:` prefix for module names where missing


thornbill:
@nielsvanvelzen there is a merge conflict here now.

PR #52, opened 2020-05-01
Update header of README

nielsvanvelzen:
Consistency!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,48 @@
-<h1 align="center">jellyfin-apiclient-java</h1>
-<h3 align="center">Part of the <a href="https://jellyfin.media">Jellyfin Project</a></h3>
+<h1 align="center">Jellyfin</h1>
+<h3 align="center">The Free Software Media System</h3>
+
+---
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,15 +1,48 @@
...
+</a>
+<a href="https://github.com/jellyfin/jellyfin-apiclient-java/commits/master.atom">
+<img alt="Master Commits RSS Feed" src="https://img.shields.io/badge/rss-commits-ffa500?logo=rss" />
+</a>
 </p>
 
 ---
 
-This library allows Android clients to easily access the Jellyfin API. It is built with Volley, OkHttp, Boon, and Robolectric. The dependencies are modular and can easily be swapped out with alternate implementations when desired.
+This library allows Java and Android applications to easily access the Jellyfin API. It is built with Volley, OkHttp, Boon, and Robolectric. The dependencies are modular and can easily be swapped out with alternate implementations when desired.
```
thornbill, line 45: This list of dependencies seem to be completely inaccurate (we only use Volley out of that list). Should we just remove that part?

nielsvanvelzen: I was going to update that part after refactoring some code. I'm looking into replacing volley too.

thornbill: I have a local implementation that uses OkHttp instead but the issue I encountered was ATV needs refactored to allow http requests to run on a different thread.

nielsvanvelzen: I'll take that in mind. When 0.7.0 gets released I will update the ATV app too

PR #53, opened 2020-05-01
Add two missing methods to improve playlist support

dkanada:


library/src/main/java/org/jellyfin/apiclient/interaction/ApiClient.java
```diff
@@ -2108,6 +2114,12 @@ public void GetPlaylistItems(PlaylistItemQuery query, final Response<ItemsResult
         GetItemsFromUrl(url, response);
     }
 
+    public void MoveWithinPlaylist(String playlist, String item, int position, final EmptyResponse response) {
```
nielsvanvelzen, line 2117: This function is called "MoveItem" in the swagger spec, I think almost all functions use those names so it should be changed to stay consistent.

PR #55, opened 2020-05-02
Refactor logging

nielsvanvelzen:
The original intention was to add Kotlin to the Android module and rewrite the AndroidLogger. In the end the following changes were made:

- ILogger and NullLogger moved to library (from model)
- Removed ILoggerManager and LogSeverity (unused)
- Reduced amount of logging levels to:
  - debug
  - info
  - error
  - exception
- Rewrite AndroidLogger in Kotlin
- ILogger now uses camelCase instead of PascalCase for methods


thornbill:
I think I would prefer the levels to be debug, info, warn, and error and allow optional passing of exceptions in errors.

nielsvanvelzen:
I chose the current values as those were the ones actually used in the apiclient itself, the others were never used.

thornbill:
Okay cool. I have no objections to leaving out `warn` until it is needed then (if it ever is). I would like to see `error` and `exception` combined into a single `error` function with an optional parameter for the throwable though. The `ErrorException` method that was replaced logged at the `error` level.

android/src/main/java/org/jellyfin/apiclient/logging/AndroidLogger.kt
```diff
@@ -0,0 +1,30 @@
...
+import android.util.Log
+
+/**
+ * ILogger implementation for Android. Uses the [android.util.Log] class.
+ */
+class AndroidLogger(
+	private val tag: String = "jellyfin-apiclient"
+) : ILogger {
+	private fun getFormattedString(raw: String, vararg parameters: Any): String {
+		return String.format(raw, *parameters)
```
thornbill, line 12: I believe @AndreasGB had to make some changes to the initial logging class to avoid calling `String.format` when no parameters are present because it would cause an exception if the string contained a %. This doesn't seem to cover that case unless I am missing some Kotlin magic. :wink:

nielsvanvelzen: As far as I could see all logging statements in the apiclient did not use string concatenation to add variables and didn't use the `%` character except for using parameters so it _should_ be safe without the check now as ATV moved to it's own logger too.

AndreasGB: I'd still program a bit defensively here to avoid shooting ourselves in the foot in the future.

PR #61, opened 2020-05-13
Rewrite server discovery

nielsvanvelzen:
Depends on #60 

This PR rewrites the server discovery logic. It doesn't integrate that nicely into the converted C# code yet.

Other changes:
- Always use the gson serializer (in the end GSON should be our only JSON parser)
- deprecated the server storage (should be removed entirely)
- misc changes


android/src/main/java/org/jellyfin/apiclient/interaction/AndroidConnectionManager.java
```diff
@@ -3,31 +3,27 @@
...
-import org.jellyfin.apiclient.serialization.IJsonSerializer;
-
-import java.util.ArrayList;
 
 public class AndroidConnectionManager extends ConnectionManager {
 
-    public AndroidConnectionManager(Context context, IJsonSerializer jsonSerializer, ILogger logger, IAsyncHttpClient httpClient, String applicationName, String applicationVersion, IDevice device, ClientCapabilities clientCapabilities, ApiEventListener apiEventListener) {
-        super(new AndroidCredentialProvider(jsonSerializer, context, logger),
-                jsonSerializer,
+    public AndroidConnectionManager(Context context, ILogger logger, IAsyncHttpClient httpClient, String applicationName, String applicationVersion, IDevice device, ClientCapabilities clientCapabilities, ApiEventListener apiEventListener) {
```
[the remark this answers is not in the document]
nielsvanvelzen: Well it was 9 so /shrug

thornbill:
Wah wah now there are conflicts.

library/src/main/java/org/jellyfin/apiclient/serialization/GsonJsonSerializer.java
```diff
@@ -1,14 +1,14 @@
...
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 
+import org.jellyfin.apiclient.model.apiclient.ServerInfo;
+
 import java.io.InputStream;
 
 public class GsonJsonSerializer implements IJsonSerializer {
-    private final Gson gson;
+    public final Gson gson;
```
thornbill, line 11: Ewww public property! :nauseated_face:

android/build.gradle.kts
```diff
@@ -30,8 +39,12 @@ android {
 dependencies {
 	api(project(":library"))
 
-	implementation(kotlin("stdlib-jdk7"))
+	implementation(kotlin("stdlib-jdk8"))
+
+	implementation("androidx.core:core-ktx:1.2.0")
+	implementation("androidx.annotation:annotation:1.1.0")
 
+	implementation("com.google.code.gson:gson:2.8.6")
```
thornbill, line 47: Why do we need `gson` here?

library/src/main/java/org/jellyfin/apiclient/discovery/ServerDiscovery.kt
```diff
@@ -0,0 +1,100 @@
...
+import com.google.gson.Gson
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.isActive
+import org.jellyfin.apiclient.logging.ILogger
+import java.net.DatagramPacket
+import java.net.DatagramSocket
+import java.net.InetAddress
+
+class ServerDiscovery(
```
thornbill, line 12: I'm guessing it will be up to clients now to create an instance of this class and use it prior to initializing the `ConnectionManager`? Could we document the usage somewhere?

library/src/main/java/org/jellyfin/apiclient/serialization/GsonJsonSerializer.java
nielsvanvelzen: It looks ugly right now but the end goal with this change it to use Gson directly and deprecate the `IJsonSerializer` interface. This way we can test everything with a single json serializer (gson) to make sure no issues occur.
Because I use Gson directly in some places right now I made it a public property

android/build.gradle.kts
nielsvanvelzen: We don't need it, it's a leftover from a previous commit. whoops

library/src/main/java/org/jellyfin/apiclient/discovery/ServerDiscovery.kt
nielsvanvelzen: I want to add a new class that is simply called `Jellyfin` which contains an instance of this class. The idea is to always use the `Jellyfin` class as entrypoint to the apiclient functionality.

A simple example will be something like this;
```kt
val jf = Jellyfin() // Optionally add options to the constructor
val flow = jf.serverDiscovery.discover(timeout = 30.seconds)

// And some future ideas:
val server = jf.server("https://demo.jellyfin.org/stable")
val info = server.getPublicInfo()

val user = server.user("token")
val user = server.authenticate("username", "password")
```

I'm still tinkering about that though, I want a solution that is compatible with OpenAPI generated code

PR #67, opened 2020-06-14
Add support for SearchTerm on ItemByNameQueries

bendardenne:
Add support for SearchTerm on by-name API queries.
See [backend](https://github.com/jellyfin/jellyfin/blob/18953d95e6692290bf56547d4c9614790687654c/MediaBrowser.Api/UserLibrary/BaseItemsByNameService.cs#L136).

### DIFF
--- a/model/src/main/java/org/jellyfin/apiclient/model/querying/ItemsByNameQuery.java
+++ b/model/src/main/java/org/jellyfin/apiclient/model/querying/ItemsByNameQuery.java
@@ -240,6 +240,20 @@ public class ItemsByNameQuery
 		NameLessThan = value;
 	}
 
+	/**
+	 Search characters used to find items
+	 <value>The index by.</value>
+	 */
+	private String SearchTerm;
+	public final String getSearchTerm()
+	{
+		return SearchTerm;
+	}
+	public final void setSearchTerm(String value)
+	{
+		SearchTerm = value;
+	}
+
 	/** 
 	 Gets or sets a value indicating whether this instance is played.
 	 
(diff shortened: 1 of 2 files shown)


model/src/main/java/org/jellyfin/apiclient/model/querying/ItemsByNameQuery.java
```diff
@@ -240,6 +240,20 @@ public final void setNameLessThan(String value)
 		NameLessThan = value;
 	}
 
+	/**
+	 Search characters used to find items
+	 <value>The index by.</value>
```
nielsvanvelzen, line 245: Most of this file was generated quite some time ago and  the javadoc it produced isn't exactly correct. You can remove this `<value>...` line as it doesn't do anything.

bendardenne: I figured this looked like generated code, but didn't see any info on how it was generated etc. So I just copy/pasted from ItemQuery.

If you want I could also reformat the whole file and get the static analyzer happy about those fields that should be at the top of the class, if this file is not going to be re-generated again.

dkanada: The long term goal is generating most of this with OpenAPI but we don't have an exact estimate on when that should be available. The migration is currently in progress over on the server.

nielsvanvelzen: We're working on adding (proper) OpenAPI support to the server so we can generate everything properly so this file will be regenerated at some point. Just add/fix what's needed now and don't care too much about all other code.

bendardenne: Okay. For now all I need is to expose this extra parameter.

PR #70, opened 2020-07-20
Add kotlin-console example using kotlinx.cli and the library

nielsvanvelzen:
# Kotlin Console Example

This sample project uses the kotlinx-cli library to build a command line tool that uses the Jellyfin
library. It's used as a showcase of the libraries abilities and is not meant for general use.

Features include:
- Server discovery
- Authenticate
- List libraries

## Basic usage
Assuming the binary is called `jellyfin` the following sample will list all libraries in the demo instance:
```sh
jellyfin libraries --server https://demo.jellyfin.org/stable --token $(jellyfin login --server https://demo.jellyfin.org/stable --username demo)
```

This command is also provided in the `test.sh` file. It requires a local install first using `./gradlew installDist`.

---

That was a copy of the README from the sample. This sample is purely meant for testing the library for now. It only contains some really basic functions right now but it can easily be extended in the future.
The command mentioned in the README and the test.sh file actually works and will print:
```
Playlists
Movies
Music
TV Shows
```

### DIFF
--- a/samples/kotlin-console/src/main/kotlin/org/jellyfin/sample/console/utils/GarbageHttpClient.kt
+++ b/samples/kotlin-console/src/main/kotlin/org/jellyfin/sample/console/utils/GarbageHttpClient.kt
@@ -1,0 +1,48 @@
+package org.jellyfin.sample.console.utils
+
+import org.jellyfin.apiclient.interaction.Response
+import org.jellyfin.apiclient.interaction.http.HttpRequest
+import org.jellyfin.apiclient.interaction.http.IAsyncHttpClient
+import java.net.HttpURLConnection
+import java.net.URL
+
+class GarbageHttpClient : IAsyncHttpClient {
+	override fun Send(request: HttpRequest?, response: Response<String>?) {
+		requireNotNull(request)
+		requireNotNull(response)
+
+		val headers = request.requestHeaders.toMap().toMutableMap()
+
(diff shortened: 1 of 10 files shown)


nielsvanvelzen:
Some future suggestions:
- Add the server & token to environment variables so executing commands doesn't need the `--server` and `--token` options
- Add exit codes and stderr support for the commands
- Add more commands
- Add output type like JSON for more data

samples/kotlin-console/src/main/kotlin/org/jellyfin/sample/console/utils/GarbageHttpClient.kt
```diff
@@ -0,0 +1,48 @@
+package org.jellyfin.sample.console.utils
+
+import org.jellyfin.apiclient.interaction.Response
+import org.jellyfin.apiclient.interaction.http.HttpRequest
+import org.jellyfin.apiclient.interaction.http.IAsyncHttpClient
+import java.net.HttpURLConnection
+import java.net.URL
+
+class GarbageHttpClient : IAsyncHttpClient {
```
thornbill, line 9: 🤣

nielsvanvelzen: I mean, it's not wrong :p

PR #74, opened 2020-08-02
Remove more sync code and remove unused models

nielsvanvelzen:
- Removed more sync code
- Removed ApiHelpers (only contained one static function that was used in a single place)
- Removed a bunch of unused models
- Removed `IHasProviderIds` interface as it was only used by `BaseItemDto`


model/src/main/java/org/jellyfin/apiclient/model/dto/BaseItemDto.java
```diff
@@ -17,15 +16,14 @@
...
-import org.jellyfin.apiclient.model.sync.SyncJobItemStatus;
 
 /** 
  This is strictly used as a data transfer object from the api layer.
  This holds information about a BaseItem in a format that is convenient for the client.
 */
 //C# TO JAVA CONVERTER TODO TASK: Java annotations will not correspond to .NET attributes:
 //ORIGINAL LINE: [DebuggerDisplay("Name = {Name}, ID = {Id}, Type = {Type}")] public class BaseItemDto : IHasProviderIds, IItemDto, IHasServerId, IHasSyncInfo
-public class BaseItemDto implements IHasProviderIds
+public class BaseItemDto
```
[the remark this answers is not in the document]
nielsvanvelzen: Not my fault!

model/src/main/java/org/jellyfin/apiclient/model/dlna/DeviceProfile.java
```diff
@@ -517,7 +517,7 @@ public final TranscodingProfile GetVideoTranscodingProfile(String container, Str
 			}
 
 			String tempVar = i.getVideoCodec();
-			if (!StringHelper.equalsIgnoreCase(videoCodec, (tempVar != null) ? tempVar : ""))
+			if(!videoCodec.equalsIgnoreCase((tempVar != null) ? tempVar : ""))
```
thornbill, line 520: This is no longer null safe and I don't see any checks to prevent `videoCodec` being null here.

model/src/main/java/org/jellyfin/apiclient/model/dlna/ConditionProcessor.java
```diff
@@ -125,9 +125,9 @@ private boolean IsConditionSatisfied(ProfileCondition condition, String currentV
 			case EqualsAny:
 				return ListHelper.ContainsIgnoreCase(expected.split("[|]", -1), currentValue);
 			case Equals:
-				return StringHelper.equalsIgnoreCase(currentValue, expected);
+				return currentValue.equalsIgnoreCase(expected);
```
thornbill, line 128: There is a check above to ensure `currentValue` is not null, so this should be fine.

model/src/main/java/org/jellyfin/apiclient/model/dlna/DeviceProfile.java
```diff
@@ -517,7 +517,7 @@ public final TranscodingProfile GetVideoTranscodingProfile(String container, Str
 			}
 
 			String tempVar = i.getVideoCodec();
-			if (!StringHelper.equalsIgnoreCase(videoCodec, (tempVar != null) ? tempVar : ""))
+			if(videoCodec != null && !videoCodec.equalsIgnoreCase((tempVar != null) ? tempVar : ""))
```
thornbill, line 520: I _think_ we want to `continue` here if `videoCodec` is `null`. Based on how the `equalsIgnoreCase` function worked, `tempVar` would never be null, and if `videoCodec` was null it would return `false` and then be inverted.

nielsvanvelzen: Yeah that was a mistake on my part trying to fix it quickly. Should be good now!

PR #76, opened 2020-08-07
Use JSON instead of forms (master branch)

nielsvanvelzen:
Confirmed working with sample client auth


nielsvanvelzen:
I accidentally pushed it to the release branch [diff](https://github.com/jellyfin/jellyfin-apiclient-java/compare/v0.6.3...1f06f4077bb4bfbf08873746762e1cb78c9a4b0c).
It's the same change (minus the garbage client) with the android version upped.

PR #78, opened 2020-08-08
Rewrite Azure pipeline based on ATV app pipeline

### LINKED ISSUE #59: Publishing automation
When #54 is merged we can publish the apiclient to Bintray which is _nice_. What would be *really* nice however is this process being automated.

We should add a new task to the Azure pipeline that will run the `publishDefaultPublicationToBintrayRepository` Gradle job with the environment variables specified in #54 set. This way we can automagically publish to Bintray.

Another thing that we could probably do (however, I did not test it) is publishing the master branch with the version set as `SNAPSHOT`.

nielsvanvelzen:
# ⭐  Everything in this PR is untested! 🚀 🌟 

What this should do:
- Build master and PR's
- Test master and PR's
- Upload .jar files to GitHub releases (including source jars)
- Publish library to [Bintray](https://bintray.com/jellyfin/jellyfin-apiclient-java/jellyfin-apiclient-java)
- Set version of the library based on the git tag

What needs to be done by someone with access to Azure:
- Add 2 environment variables called `BINTRAY_USER` and `BINTRAY_KEY` with credentials for the publisher. We should probably create a bot account for that.

If everything is ready and checked by living humans with more Azure pipeline knowledge than me we could theoretically release 0.7.0.

---

See #54 for the Bintray PR

---

This PR closes #59

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -81,4 +81,4 @@ jobs:
           assets: '**/libs/*.jar'
           action: 'edit'
           assetUploadMode: 'replace'
-          tag: '$(TAG)'
+          tag: '$(JELLYFIN_VERSION)'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,15 +1,84 @@
+variables:
+  - group: 'jellyfin'
+
 trigger:
-- master
+  batch: true
+  branches:
+    include:
+      - master
+      - release*
+  tags:
+    include:
+      - '*'
 
-pool:
-  vmImage: 'ubuntu-latest'
+pr:
+  branches:
+    include:
+      - '*'
 
-steps:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -1,15 +1,84 @@
...
+
+      - task: GithubRelease@0
+        displayName: 'GitHub Upload'
+        inputs:
+          gitHubConnection: Jellyfin Release Download
+          repositoryName: jellyfin/jellyfin-apiclient-java
+          assets: '**/libs/*.jar'
+          action: 'edit'
+          assetUploadMode: 'replace'
+          tag: '$(TAG)'
```
thornbill, line 84: Looks like you used `JELLYFIN_VERSION` for this variable name above.

PR #81, opened 2020-08-15
Add buildSrc gradle project

nielsvanvelzen:
This will solve some things:
- No need to paste functions inside of build.gradle.kts scripts, making those more readable
- Sharing dependency versions, update them in a single place
- Slightly improved build times (barely noticable)

Not a lot of code was added, it was mostly moving things around. Some new code in `Dependencies.kt` though.

### DIFF
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -1,0 +1,44 @@
+import org.gradle.api.artifacts.dsl.DependencyHandler
+import org.gradle.kotlin.dsl.kotlin
+import org.gradle.kotlin.dsl.project
+
+object Dependencies {
+	object KotlinX {
+		private fun item(module: String, version: String) = "org.jetbrains.kotlinx:kotlinx-${module}:${version}"
+
+		val cli = item("cli", "0.2.1")
+		val coroutinesCore = item("coroutines-core", "1.3.5")
+	}
+
+	object AndroidX {
+		private fun item(library: String, module: String = library, version: String) = "androidx.${library}:${module}:${version}"
+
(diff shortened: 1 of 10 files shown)


buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -0,0 +1,44 @@
...
+ * Add given project as implementation
+ */
+fun DependencyHandler.implementationProject(path: String) {
+	add("implementation", project(path))
+}
+
+/**
+ * Add the Kotlin standard library as implementation
+ */
+fun DependencyHandler.implementationKotlinStdlib() = kotlin("stdlib")
```
Maxr1998, line 44: If I'm not mistaken, this is missing a surrounding `implementation()`

nielsvanvelzen: Good catch!

buildSrc/src/main/kotlin/Semver.kt
```diff
@@ -0,0 +1,26 @@
...
+ * 0.0.0 -> 0
+ * 1.1.1 -> 10101
+ * 0.7.0 -> 700
+ * 99.99.99 -> 999999
+ */
+fun getVersionCode(semver: String): Int? {
+	val parts = semver.splitToSequence('.')
+		.take(3)
+		.map { it.toIntOrNull() }
+		.filterNotNull()
```
Maxr1998, line 14: I think you could use `mapNotNull` here.

nielsvanvelzen: Yup, IDE didn't hint it and I didn't think about it when writing the function, changed it now.

PR #83, opened 2020-08-15
Fix release assets uploading the wrong files

nielsvanvelzen:
I did some testing and noticed that more .jar files were be picked up (buildSrc.jar). I made the glob more strict to only upload the generated libraries.
I also added distributions to the uploads, this currently includes a build of the console example (called "kotlin-console-$VERSION.zip") that includes all used libraries etc. for easy use.

The globs are made in a way that adding new samples or library modules won't require changes.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -81,7 +81,7 @@ jobs:
           gitHubConnection: Jellyfin Release Download
           repositoryName: jellyfin/jellyfin-apiclient-java
           assets: |
-            "**/distributions/*-${JELLYFIN_VERSION}.zip
+            **/distributions/*-${JELLYFIN_VERSION}.zip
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -80,7 +80,10 @@ jobs:
         inputs:
           gitHubConnection: Jellyfin Release Download
           repositoryName: jellyfin/jellyfin-apiclient-java
-          assets: '**/libs/*.jar'
+          assets: |
+            "**/distributions/*-${JELLYFIN_VERSION}.zip
```
Maxr1998, line 84:
```suggestion
            **/distributions/*-${JELLYFIN_VERSION}.zip
```

nielsvanvelzen: _bad copy paste job_

PR #84, opened 2020-08-17
Fix Java files not being included in the Android build

nielsvanvelzen:
Somehow Gradle cached my build so I did not catch this issue when testing #80. The .java files were not included in builds. I moved the java files back to the java source set for now.


Maxr1998:
I had the same happen to me the other day :smile:

PR #90, opened 2020-08-30
add method to get image blurhashes

dkanada:


Maxr1998 (approved):

nielsvanvelzen:
This should've targeted master and cherrypicked into the release afterwards. Also, there is a ImageBlurHashes on BaseItemPerson too which was not included in this PR. Do you need that too or not?

dkanada:
Hmmmm I don't believe so, but album artists might be people depending on which endpoint you get them from. I use the one that should return a base item so I think it's fine.

PR #93, opened 2020-09-06
Update model module to support generated sources

nielsvanvelzen:
Part **2** of multiple

Merge after #92

### DIFF
--- a/model/build.gradle.kts
+++ b/model/build.gradle.kts
@@ -2,6 +2,12 @@ plugins {
 	id("kotlin")
 }
 
+dependencies {
+	implementation(kotlin("stdlib-jdk7"))
+}
+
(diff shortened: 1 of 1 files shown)


model/build.gradle.kts
```diff
@@ -2,14 +2,20 @@ plugins {
 	id("kotlin")
 }
 
+dependencies {
+	implementation(kotlin("stdlib-jdk7"))
```
Maxr1998, line 6: Why not stdlib-jdk8?

Maxr1998 (approved):

model/build.gradle.kts
nielsvanvelzen: I think I forgot to change it in this specific file. Will be fixed in another PR as I want to migrate everything to buildSrc

PR #94, opened 2020-09-06
Add minimal API module with hardcoded Authorization header

nielsvanvelzen:
Part **3** of multiple

Merge after #92 

The current implementation is minimal and some code is missing (extensions + java compat). API calls use Kotlin delegates to make pretty code. Calling an API will look like this:

```kotlin
fun `authentication test`() = runBlocking {
	val userApi = UserApi(...)

	val auth: Response<AuthenticationResult> = userApi.authenticateUserByName("demo", "")
	val authDelegated: AuthenticationResult by userApi.authenticateUserByName("demo", "")
}
```

### DIFF
--- a/api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
+++ b/api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
@@ -1,0 +1,141 @@
+package org.jellyfin.apiclient.api.client
+
+import com.google.gson.FieldNamingPolicy
+import io.ktor.client.*
+import io.ktor.client.call.*
+import io.ktor.client.features.json.*
+import io.ktor.client.request.*
+import io.ktor.client.statement.*
+import io.ktor.http.*
+import org.jellyfin.apiclient.api.client.adapter.LocalDateTimeTypeAdapter
+import org.jellyfin.apiclient.api.client.adapter.UUIDTypeAdapter
+import java.time.LocalDateTime
+import java.util.*
+
+open class KtorClient(
+	var baseUrl: String
+) {
+	val client = HttpClient {
+		install(JsonFeature) {
+			serializer = GsonSerializer {
+				setFieldNamingPolicy(FieldNamingPolicy.UPPER_CAMEL_CASE)
+				serializeNulls()
+				registerTypeAdapter(UUID::class.java, UUIDTypeAdapter())
+				registerTypeAdapter(LocalDateTime::class.java, LocalDateTimeTypeAdapter())
+			}
+		}
+	}
+
+	/**
+	 * Replaces the variables in a path with the values in the map.
+	 * Will also remove trailing and repeated slashes
+	 */
+	fun createPath(path: String, pathParameters: Map<String, Any?>) = path
+		.split('/')
+		.filterNot { it.isEmpty() }
+		.map { rawName ->
+			val name = rawName.removeSurrounding(prefix = "{", suffix = "}")
+
+			// only act if the name actually is a variable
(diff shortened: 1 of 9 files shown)


api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
```diff
@@ -0,0 +1,141 @@
...
+	var baseUrl: String
+) {
+	val client = HttpClient {
+		install(JsonFeature) {
+			serializer = GsonSerializer {
+				setFieldNamingPolicy(FieldNamingPolicy.UPPER_CAMEL_CASE)
+				serializeNulls()
+				registerTypeAdapter(UUID::class.java, UUIDTypeAdapter())
+				registerTypeAdapter(LocalDateTime::class.java, LocalDateTimeTypeAdapter())
+			}
```
Maxr1998, line 25: I've recently read Gson shouldn't be used anymore, maybe we should switch to something more modern and Kotlin-aware like kotlinx.serialization for the new client?

api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	suspend inline fun <reified T> get(
+		pathTemplate: String,
+		pathParameters: Map<String, Any?> = emptyMap(),
+		queryParameters: Map<String, Any?> = emptyMap(),
+		body: Any? = null
+	) = request<T>(
+		method = HttpMethod.Get,
+		path = createPath(pathTemplate, pathParameters),
```
Maxr1998, line 113: I think you could move the `createPath` invocation to the `request` function as well.

Maxr1998 (approved):

api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
nielsvanvelzen: kotlinx.serialization requires an annotation AFAIK but the `model` module doesn't have external dependencies and I want to keep it that way.
Do you have a source about not using gson, I'm quite intested in that.

api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
nielsvanvelzen: I didn't do that to make the request function slightly more simple I think, I'll have another look at it later.

api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
Maxr1998: Yep, kotlinx.serialization does require an annotation unfortunately. [Here's a comment](https://www.reddit.com/r/Kotlin/comments/exmp2s/json_to_kotlin_data_class/fgak5mc/) from Jake Wharton on the state of Gson from already a few months ago. It was recently references in [this](https://www.reddit.com/r/Kotlin/comments/ikwecd/whats_the_best_json_to_kotlin_objects_converter/) thread.

nielsvanvelzen: Ah thanks, I might switch to kotlinx.serialization in a future PR. I'd love to make the apiclient a multiplatform project but that will be quite some work because simple things like UUID's are not in Kotlin's stdlib and there is no date/time implementation yet.

Maxr1998 (approved):

PR #96, opened 2020-09-06
Add UserApi.authenticateUserByName extension

nielsvanvelzen:
Part **5** of multiple

Merge after #94

Currently just a single extension but the idea is to add more in the future. These extensions help simplifying API usage.

```kotlin
// Before
userApi.authenticateUserByName(AuthenticateUserByName(username = "demo", pw = null, password = "demo"))
// After
userApi.authenticateUserByName(username = "demo", password = "demo")
```

### DIFF
--- a/api/src/main/kotlin/org/jellyfin/apiclient/api/client/extensions/UserApiExtensions.kt
+++ b/api/src/main/kotlin/org/jellyfin/apiclient/api/client/extensions/UserApiExtensions.kt
@@ -1,0 +1,14 @@
+package org.jellyfin.apiclient.api.client.extensions
+
+import org.jellyfin.apiclient.api.operations.UserApi
+import org.jellyfin.apiclient.model.AuthenticateUserByName
+
+/**
+ * Extension function for the authenticateUserByName operation that accepts the username and password directly
+ */
+suspend inline fun UserApi.authenticateUserByName(username: String, password: String) = authenticateUserByName(
+	data = AuthenticateUserByName(
+		username = username,
+		password = password
+	)
+)


api/src/main/kotlin/org/jellyfin/apiclient/api/client/extensions/UserApiExtensions.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.jellyfin.apiclient.model.AuthenticateUserByName
+
+/**
+ * Extension function for the authenticateUserByName operation that accepts the username and password directly
+ */
+suspend inline fun UserApi.authenticateUserByName(username: String, password: String) = authenticateUserByName(
+	data = AuthenticateUserByName(
+		username = username,
+		password = password
+	)
```
Maxr1998, line 13: Not sure whether you have to specify every parameter name directly, but ok ^^

Maxr1998 (approved):

api/src/main/kotlin/org/jellyfin/apiclient/api/client/extensions/UserApiExtensions.kt
nielsvanvelzen: The reason I did that is because `AuthenticateUserByName` is generated by the generator and if it (for some reason) changes the compiler will complain to us.
It also skips the `pw` property (because yes we have password and pw) which if I remember correctly is the second parameter, defaulting to null.

Maxr1998: Ah, that makes sense then. I remember that pw thing from talks inside the chat a few days ago.

thornbill (approved):

PR #97, opened 2020-09-09
Use buildSrc in new modules

nielsvanvelzen:
The buildSrc structure wasn't added yet when I started on the OpenAPI branch and updating would be a pain before everything was merged. So here it is!

Fixes the mix of stdlib-jdk7 and stdlib-jdk8.

I did not update the versions, although using Kotlin 1.4 is planned when the new kotlinpoet release is released.

### DIFF
--- a/android/build.gradle.kts
+++ b/android/build.gradle.kts
@@ -58,7 +58,7 @@ afterEvaluate {
 dependencies {
 	apiProject(":library")
 
-	implementationKotlinStdlib()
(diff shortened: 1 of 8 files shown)


nielsvanvelzen:
PS: Build fails because of the authentication extension, will be fixed when the generated code is pushed.

Maxr1998 (approved):

thornbill (approved):

PR #100, opened 2020-09-10
Remove old API code

nielsvanvelzen:
Part **7** of multiple

I fixed a few code quality issues.

Removed the old API code + models. I disabled the createApi function for now (will implement in a separate PR). I had to disable the sample too because it depends on the old API code. Next milestone is to actually be able to use the new API.


Maxr1998 (approved):

Maxr1998:
Gave it a quick read, should be good.

library/src/main/java/org/jellyfin/apiclient/IDevice.kt
```diff
@@ -1,4 +1,4 @@
-package org.jellyfin.apiclient.interaction.device
+package org.jellyfin.apiclient
 
 interface IDevice {
```
thornbill, line 3: Can we ditch the “I” prefix since we’re changing things?

nielsvanvelzen: I'm going to do that when implementing the new API. This interface might be removed entirely.

thornbill (approved):

PR #102, opened 2020-09-11
Change generated models package to org.jellyfin.apiclient.model.api

nielsvanvelzen:
While working on the API implementation I noticed the models got a conflicting name and this package name makes more sense anyway.


Maxr1998 (dismissed):

Maxr1998:
Waiting for checks first.

thornbill (approved):

PR #105, opened 2020-09-13
Update license to LGPL 3.0

nielsvanvelzen:
While the MIT license allows the use of this library for a lot of use-cases it's too "free" in my opinion. The GPL family is a bit more strict to make the "free" code more "fair". I discussed this in our Matrix channels already. I also considered MPL but chose not to because of the file headers (I'm not sure if these are required and I don't want to make a mistake there by not using it when I should).

According to a small research I did it should be possible to sublicense a MIT project to (L)GPL 3.0 <sup>[1]</sup> <sup>[2]</sup>. And even if this wasn't allowed the current codebase is (roughly) 90% written by me since I refactored almost everything. Only some smaller files (like the gitignore or README) are still based on pre-fork content.

If we choose to make the apiclient a kotlin-multiplatform project and start supporting multiple platforms (like Apple) we need to be careful as Apple doesn't allow LGPL due too dynamic linking. I've read somewhere (sorry lost the source) that adding a notice to the README about allowing static linking could be enough to "fix" this (but I'm not a lawyer!).

I chose to use a plaintext file instead of markdown for the license. 

[1]: https://softwareengineering.stackexchange.com/questions/105912/can-you-change-code-distributed-under-the-mit-license-and-re-distribute-it-unde
[2]: https://softwareengineering.stackexchange.com/questions/189633/what-sublicense-actually-means


joshuaboniface (approved):
I think this seems sound from a licensing perspective. Sublicensing MIT as (L)GPL is OK, esp if you've rewritten most of the code. I'm not certain about the blurb method of allowing static linking, but since you'd also own the copyright you can actually grant whatever rights you want (this was only an issue with the old code since we *didn't* hold exclusive copyright to make that decision).

thornbill (approved):

oddstr13 (changes requested):
The MIT license file must be retained when sublicensing.
> The above copyright notice and this permission notice shall be included in
> all copies or substantial portions of the Software.

It can be removed only if all contributors (copyright holders) agree.

Can rename it to something along the lines of `MIT-Copytight-notice.md` however.

nielsvanvelzen:
I took a look into the code and can only find 3 Java classes that are still from before-fork: ILogger, ConsoleLogger and NullLogger. They can be easily replaced with something from scratch (although API compatible by keeping the names). All other code is written by me. This means that all files inside of directories can actually be relicensed without keeping the MIT license. (although some might have a history with pre-fork contributors due too the way we rewrote a bunch of code but kept certain file names).

The only problem I see is some files in the repository root like the README or Azure Pipeline YAML. I'm not sure if those are significant enough to claim authorship too and they are mostly (if not entirely) written by Jellyfin contributors.

Do you still think we need to keep the MIT license @oddstr13? And if so: will renaming the MIT license file to "LICENSE-ORIGINAL" be enough or do we need to mention in the README why there's multiple license files?

dkanada:
I'm pretty sure the README is new since Jellyfin started, and I believe the CI pipeline was copied from Android,  so I probably hold the copyright for that one. Can you even copyright basic things like pipelines? I thought there was some history for code that was generic enough to be considered almost universal.

oddstr13 (approved):
If you are confident that you have sufficiently replaced everything so that the original Emby copyrights are no longer a concern, I won't block the removal of the file.

I wouldn't count basic pipelines stuff sufficiently original to be copyrightable, unless they are larger scripts rather than just a few commands.

If it is just a few files that are potentially affected, just moving the MIT license into comments in those files should be sufficient.

Obligatory [IANAL](https://en.wikipedia.org/wiki/IANAL).

PR #109, opened 2020-09-17
Openapi api implementation

nielsvanvelzen:
- Finalized HTTP API
- Updates sample to new code

It works sort of, the new ping command in the CLI only works on unstable right now because of a new property that was added to public server info responses and the login command doesn't work because of an error in the OpenAPI spec, this should be fixed in the server first.


api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
```diff
@@ -37,25 +48,32 @@ open class KtorClient(
...
+			"deviceId" to deviceInfo.id,
+			"device" to deviceInfo.name
 		)
 
-//		params["Token"] = ""
+		// Only set access token when it's not null
+		accessToken?.let { params["token"] = it }
 
-		// Format: `MediaBrowser key="value", key="value"`
+		// Format: `MediaBrowser key1="value1", key="value"`
```
Maxr1998, line 62:
```suggestion
		// Format: `MediaBrowser key="value", key1="value1"`
```

Maxr1998 (approved):

api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
nielsvanvelzen: I swear I used key1=value1 key2=value2 at some point, no idea what happened to it.

Maxr1998 (approved):

thornbill (approved):

PR #110, opened 2020-09-20
Nullable references

nielsvanvelzen:
This PR adds a hook that makes references to objects in models nullable as workaround for the missing nullable properties in the OpenAPI document.

Ideally this should be fixed in server but with this workaround I can continue developing the apiclient for now.


Maxr1998 (approved):
LGTM

PR #112, opened 2020-09-26
Add operation url function to API services

nielsvanvelzen:
Because we need things like `ImageApi.getItemImageUrl` for Glide etc.

It uses a hook so we can easily change the behavior. The current behavior is to always create url functions for operations that use a GET request and return binary data.

_Like always, I suggest reviewing per commit._

### DIFF
--- a/api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
+++ b/api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
@@ -17,7 +17,7 @@ open class KtorClient(
 	var accessToken: String? = null,
 	var clientInfo: ClientInfo,
 	var deviceInfo: DeviceInfo
-): ApiClient {
+) : ApiClient {
 	val client = HttpClient {
 		install(JsonFeature) {
 			serializer = KotlinxSerializer()
@@ -104,15 +104,15 @@ open class KtorClient(
 		pathTemplate: String,
 		pathParameters: Map<String, Any?> = emptyMap(),
 		queryParameters: Map<String, Any?> = emptyMap(),
-		body: Any? = null
+		requestBody: Any? = null
 	): Response<T> {
 		val response = client.request<HttpResponse> {
 			this.method = method
 			url(createUrl(pathTemplate, pathParameters, queryParameters))
 			header(HttpHeaders.Authorization, createAuthorizationHeader())
 
-			if (body != null)
-				this.body = defaultSerializer().write(body)
+			if (requestBody != null)
+				body = defaultSerializer().write(requestBody)
 		}
 
 		return Response(response.receive())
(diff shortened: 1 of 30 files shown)


nielsvanvelzen:
Some additional notes:
- I started on using an interface for the apiclient so we can, for example, create an implementation to mock the API. I didn't finish it though because inline functions don't work in interfaces so I need to think of a solution for that
- I also thought about using the `Response` class (and renaming it to Request) so you can do something like:
  ```kotlin
  // get url
  val url = ImageApi(api).getItemImage(xxx).url
  // get response
  val response by ImageApi(api).getItemImage(xxx)
  ```
  But this didn't work as the `getValue()` function needed to be suspended for that and that's impossible.

api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
```diff
@@ -84,34 +106,13 @@ open class KtorClient(
...
-						parameters.append(it.key, it.value.toString())
-					}
-			}
-
+			url(createUrl(pathTemplate, pathParameters, queryParameters))
 			header(HttpHeaders.Authorization, createAuthorizationHeader())
 
-			if (body != null) {
+			if (body != null)
 				this.body = defaultSerializer().write(body)
```
Maxr1998, line 115: You should maybe rename the passed function argument to decrease possible confusion..

openapi-generator/src/main/kotlin/org/jellyfin/openapi/MainModule.kt
```diff
@@ -28,7 +29,8 @@ val mainModule = module {
 	// API
 	single { ApiNameBuilder() }
 	single { OperationBuilder(get()) }
-	single { ApiBuilder(get()) }
+	single { OperationUrlBuilder(get()) }
+	single { ApiBuilder(get(), get(), getAll()) }
```
Maxr1998, line 33: I didn't know about `getAll()` yet, pretty cool solution! :+1:

Maxr1998 (approved):

openapi-generator/README.md
```diff
@@ -16,15 +16,21 @@ OpenAPI specification changes.
...
+    A hook that can intercept the TypeBuilder which converts OpenAPI schemas to Kotlin types. It
+    receives the schema and a type path. The path is unique across the whole document and can be used
+    to identified specific properties. This hook is called when generating types for:
+     
+      - model properties
+      - api operation parameters
+      - api operation bodies
+      - api operation return types
+
+  - **OperationUrlHook**
```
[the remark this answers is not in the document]
nielsvanvelzen: but you just told me to add it......................

thornbill (approved):

nielsvanvelzen:
I'll fix all markdown issues (for all the markdown files) in a separate PR.

Maxr1998 (approved):

PR #113, opened 2020-09-27
Improve discovery code

nielsvanvelzen:
I started on a small change and it ended up being a bit more in a single commit... whoops

  - Rewritten addressCandidates function with a lot more options
  - Added unit tests for addressCandidates
  - Changed some names to be more logical
  - Changed some kdoc to be more clear
  - Use kotlinx.serialization instead of GSON (also dropped GSON from the dependencies)
  - Remove unused Volley dependency
  - Updated logging in changed code to slf4j

See #111 for the slf4j change


library/src/test/kotlin/org/jellyfin/apiclient/discovery/DiscoveryServiceTests.kt
```diff
@@ -0,0 +1,59 @@
...
+
+		assert(instance.getAddressCandidates("demo.jellyfin.org:433/stable/").first().startsWith("https://"))
+		assert(instance.getAddressCandidates("http://demo.jellyfin.org:433/stable/").first().startsWith("https://"))
+	}
+
+	@Test
+	fun `getAddressCandidates adds Jellyfin ports`() {
+		val instance = getInstance()
+
+		assert(instance.getAddressCandidates("localhost").contains("http://localhost:8096"))
```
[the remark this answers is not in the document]
nielsvanvelzen: But it's a test...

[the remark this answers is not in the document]
nielsvanvelzen: It's a unit test stupid bot

library/src/main/java/org/jellyfin/apiclient/discovery/AddressCandidateHelper.kt
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	/**
+	 * Return all candidate URLs as string
+	 */
+	fun getCandidates(): Collection<String> = candidates
+		// Convert to strings
+		.map { it.toString() }
+		// Remove duplicates
+		.distinct()
```
Maxr1998, line 158:
```suggestion
	/**
	 * Get all deduplicated candidate URLs as strings
	 */
	fun getCandidates(): Collection<String> = candidates.map(Url::toString).distinct()
```

library/src/main/java/org/jellyfin/apiclient/discovery/DiscoveryService.kt
```diff
@@ -0,0 +1,41 @@
...
+	 *   - Add protocol (https and http)
+	 *   - Add port (8096 for http and 8920 for https)
+	 *   - Add baseurl (/jellyfin)
+	 *
+	 * The returned candidates are in order of most secure followed by most likely to work.
+	 * See [AddressCandidateHelper] for more information.
+	 */
+	fun getAddressCandidates(input: String) = AddressCandidateHelper(input).run {
+		addCommonCandidates()
+		prioritize()
```
Maxr1998, line 26: Maybe put this into the `getCandidates` function even?

nielsvanvelzen: I explicitly did no do that if an unsorted list is desired.

nielsvanvelzen: But I added an additional comment to the `getCandidates` function

Maxr1998: > I explicitly did no do that if an unsorted list is desired.

Hmm, I see. You have a typo in there though, I think.

nielsvanvelzen: You're right. I shouldn't work on 5 things at once lol

thornbill (approved):
Changes look good and I verified discovery worked using the cli sample app.

PR #116, opened 2020-10-02
Use kasechange in openapi-generator

nielsvanvelzen:
It parses some names slightly different then my solution did. For some names (like X64) this works better but for others (like is4K) it's worse.

I'm not sure if I like the previous names more then the current so I'd like to have some feedback first before merge.


Maxr1998 (dismissed):
I prefer the new naming scheme in most cases, especially removing the underscore is a very good thing. `is4k` looks fine to me, as does `is3d` on its own, only `is3dFormat` looks a bit weird imo, but at least it's consistent.

nielsvanvelzen:
Whoops, forgot to remove the unit tests for the textcase class. Will do that tomorrow.

Maxr1998 (approved):

Maxr1998:
Should we wait for a review from @thornbill or can I just merge?

nielsvanvelzen:
It's not too  big change but I'd still like @thornbill's opinion on it so let's wait a bit 😄

thornbill (approved):
:+1: LGTM

PR #119, opened 2020-10-06
Catch IO exceptions when sending discovery messages

nielsvanvelzen:
This fixes an issue reporting in our Matrix channels where the new Android app crashes on https://grapheneos.org. If approved I'll backport the try-catch statement to 0.7.x (I'll push it to the release branch).


thornbill (approved):

nielsvanvelzen:
I made a build of the Android app with this fix added and just got a confirmation that it indeed fixes the crash :)

nielsvanvelzen:
I've pushed the fix to the release branch.

Issue #107, opened 2020-09-17
Update the README


nielsvanvelzen:
- Some badges don't work
- The sample code is outdated.
- Should add a link to the sample(s)
- Should add a blurb about our own use in the Android / Android TV apps as example implementations

PR #120, opened 2020-10-07
Module name refactor

nielsvanvelzen:
Changes:

- All published modules are now prefixed with `jellyfin-`. Which is a common convention for Java libraries (see kotlinx, ktor or slf4j to just name a few examples).
- The "library" module is now named "jellyfin-core" (personal preference)
- The "android" module is now named "jellyfin-platform-android"
  - In the future we might introduce "jellyfin-platform-jvm" for non-android applications
- All sources in the "jellyfin-core" module are now in the "kotlin" sourceset instead of "java"
- Slightly changed the implementation/api use for our modules:
  - core exposes both api and model
  - android exposes core and thus api and model
  - api does not expose model


nielsvanvelzen:
Note: README changes not included, those will be in #107

Maxr1998 (approved):

thornbill (approved):

PR #122, opened 2020-10-10
Use coroutine based ByteReadChannel instead of synchronous InputStream

nielsvanvelzen:
I wasn't sure if the ByteReadChannel class would be exposed properly (since ktor is implementation, not api) but from my quick testing it seems like it does.

### DIFF
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -44,6 +44,7 @@ object Dependencies {
 		val serialization = item("client-serialization-jvm")
 
 		val http = item("http")
+		val io = item("io")
 	}
 
 	object Koin {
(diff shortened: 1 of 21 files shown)


Maxr1998 (approved):

thornbill (approved):
:+1: lgtm provided we are sure this doesn't introduce issues with Java interoperability as that would make @dkanada sad

nielsvanvelzen:
It's possible to convert a byte read channel to an input stream using the .toInputStream extension. In Java the code could look like this:

```java
package org.jellyfin.apiclient.api.client;

import org.jellyfin.apiclient.api.client.compatibility.JavaDataCallback;
import org.jellyfin.apiclient.api.operations.HlsSegmentApi;
import org.jetbrains.annotations.Nullable;

import java.io.InputStream;

import io.ktor.utils.io.ByteReadChannel;
import io.ktor.utils.io.jvm.javaio.BlockingKt;

public class DkanadaTest {
    public void test() {
        KtorClient client = new KtorClient(/*parameters*/);
        HlsSegmentApi t = new HlsSegmentApi(client);
        t.getHlsAudioSegmentLegacyAac("id", "segment", new JavaDataCallback<ByteReadChannel>() {
            @Override
            public void onData(@Nullable ByteReadChannel data) {
                InputStream stream = BlockingKt.toInputStream(data, null);
                // Do Gelli things
            }
        });
    }
}
```

PR #123, opened 2020-10-10
Add more information to the Response class

nielsvanvelzen:
Let me know if something is missing.

It uses infix functions because pretty code.

### DIFF
--- a/jellyfin-api/src/main/kotlin/org/jellyfin/apiclient/api/client/Response.kt
+++ b/jellyfin-api/src/main/kotlin/org/jellyfin/apiclient/api/client/Response.kt
@@ -24,51 +24,6 @@ class Response<T>(
 	operator fun getValue(thisRef: Any?, property: KProperty<*>): T = content
 
 	/**
-	 * Check if the HTTP response status is equal to a given status.
-	 *
-	 * ```kt
-	 * val isOk = response isStatus 200
-	 * val isOk = response.isStatus(200)
-	 * ```
-	 */
-	infix fun isStatus(comparable: Int) = status == comparable
-
-	/**
-	 * Check if the HTTP response status is not equal to a given status.
-	 *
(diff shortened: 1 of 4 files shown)


h1dden-da3m0n (approved):

jellyfin-api/src/main/kotlin/org/jellyfin/apiclient/api/client/Response.kt
```diff
@@ -1,12 +1,81 @@
...
+
+	/**
+	 * Check if the HTTP response status is in the 4xx range.
+	 */
+	fun isStatus400() = this isStatusIn (400 until 500)
+
+	/**
+	 * Check if the HTTP response status is in the 5xx range.
+	 */
+	fun isStatus500() = this isStatusIn (500 until 600)
```
Maxr1998, line 69: Are those really that necessary? I'm not sure of the benefit these functions bring..

nielsvanvelzen: I've actually used `response isStatus 200` in an upcoming PR

Maxr1998: But is that actually so much better than a `response.status == 200`?

nielsvanvelzen: Hmm.. actually you're right. The status value was private at first but since it's public anyway we don't really need these functions.

Maxr1998 (approved):

Maxr1998:
@nielsvanvelzen should we wait for a review from @thornbill or can I (squash) merge it already?

h1dden-da3m0n (approved):

PR #127, opened 2020-10-11
update(ci): change ci publish steps ... again

h1dden-da3m0n:
**Changes**

* updates how the `JELLYFIN_VERSION` gets set (uses now the same mechanism as [jellyfin/jellyfin](https://github.com/jellyfin/jellyfin/blob/2375c35c4a966a469dd97f33deb57b342498b905/.ci/azure-pipelines-package.yml#L119))
* updates GitHubRelease task to select correct tag with new `JELLYFIN_VERSION`
* change to align publish steps more with the ci in jenkins/jenkins-androidtv ([Line 75 and down](https://github.com/jellyfin/jellyfin-androidtv/blob/727a2e8e351ac9a1e18049c7dd303005ecabbff7/.ci/azure-pipelines.yml#L75))

**Issues**

* more #83 

**Relevant Documentation**

* [GitHubRelease task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/github-release)
  * > Assets - (Optional) Specify  the files to be uploaded as assets for the release. You can use wildcard  characters to specify a set of files. E.g. $(Build.ArtifactStagingDirectory)/*.zip. You can also specify multiple patterns - one per line. By default, all files in the $(Build.ArtifactStagingDirectory) directory will be uploaded.

* [CopyFiles](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/copy-files)
  * > Contents - (Required) File paths to include as part of the copy. Supports multiple lines of match patterns. More Information. For example: * copies all files in the specified source folder** copies all files in the specified source folder and all files in all sub-folders**\bin\** copies all files recursively from any bin folder The  pattern is used to match only file paths, not folder paths. So you  should specify patterns such as **\bin\** instead of **\bin.You must use the path separator that matches your build agent type. Example, / must be used for Linux agents. More examples are shown below. Default value: **

* [Build Variables (DevOps Services)](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables-devops-services)
  * > Build.ArtifactStagingDirectory - The local path on the agent where any artifacts are copied to before being pushed to their destination. For example: c:\agent_work\1\a  A typical way to use this folder is to publish your build artifacts with the Copy files and Publish build artifacts tasks.  Note: Build.ArtifactStagingDirectory and Build.StagingDirectory are  interchangeable. **This directory is purged before each new build, so you  don't have to clean it up yourself.**  See Artifacts in Azure Pipelines.  This variable is agent-scoped, and can be used as an environment  variable in a script and as a parameter in a build task, but not as part  of the build number or as a version control tag.

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -61,10 +61,11 @@ jobs:
     pool:
       vmImage: 'ubuntu-latest'
 
-    steps:
-      - script: echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$( awk -F '/' '{ print $NF }' <<<'$(Build.SourceBranch)' | sed 's/^v//' )"
-        displayName: 'Set Version Variable'
+    variables:
+      - name: JELLYFIN_VERSION
+        value: $[replace(variables['Build.SourceBranch'],'refs/tags/v','')]
 
+    steps:
       - task: Gradle@2
         displayName: 'Build, Distribute & Publish'
         inputs:
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -85,6 +85,7 @@ jobs:
             **/distributions/*-$(JELLYFIN_VERSION).zip
             **/libs/*-$(JELLYFIN_VERSION).jar
             **/libs/*-$(JELLYFIN_VERSION)-sources.jar
+            **/outputs/aar/*.aar
           TargetFolder: '$(Build.ArtifactStagingDirectory)'
           flattenFolders: true
 
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -62,7 +62,7 @@ jobs:
       vmImage: 'ubuntu-latest'
 
     steps:
-      - script: 'echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$(git describe --tags)"'
+      - script: echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$( awk -F '/' '{ print $NF }' <<<'$(Build.SourceBranch)' | sed 's/^v//' )"
         displayName: 'Set Version Variable'
 
       - task: Gradle@2
@@ -89,4 +89,4 @@ jobs:
             $(Build.ArtifactStagingDirectory)/**/libs/*-$(JELLYFIN_VERSION)-sources.jar
           action: 'edit'
           assetUploadMode: 'replace'
-          tag: '$(JELLYFIN_VERSION)'
+          tag: 'v$(JELLYFIN_VERSION)'
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -78,15 +78,21 @@ jobs:
         env:
           BINTRAY_KEY: $(bintray_key)
 
+      - task: CopyFiles@2
+        displayName: 'Copy build assets'
+        inputs:
+          Contents: |
+            **/distributions/*-$(JELLYFIN_VERSION).zip
+            **/libs/*-$(JELLYFIN_VERSION).jar
+            **/libs/*-$(JELLYFIN_VERSION)-sources.jar
+          TargetFolder: '$(Build.ArtifactStagingDirectory)'
+          flattenFolders: true
+
       - task: GitHubRelease@1
         displayName: 'GitHub Upload'
         inputs:
           gitHubConnection: Jellyfin Release Download
           repositoryName: jellyfin/jellyfin-apiclient-java
-          assets: |
-            $(Build.ArtifactStagingDirectory)/**/distributions/*-$(JELLYFIN_VERSION).zip
-            $(Build.ArtifactStagingDirectory)/**/libs/*-$(JELLYFIN_VERSION).jar
-            $(Build.ArtifactStagingDirectory)/**/libs/*-$(JELLYFIN_VERSION)-sources.jar
           action: 'edit'
           assetUploadMode: 'replace'
           tag: 'v$(JELLYFIN_VERSION)'


.ci/azure-pipelines.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       vmImage: 'ubuntu-latest'
 
     steps:
-      - script: 'echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$(git describe --tags)"'
+      - script: echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$( awk -F '/' '{ print $NF }' <<<'$(Build.SourceBranch)' | sed 's/^v//' )"
```
nielsvanvelzen, line 65: `Build.SourceBranch` implies it's the name of a branch. For the apiclient (and Android repositories) we use tags to set the version and the branch can be anything (most of the time release branches though).

nielsvanvelzen: Also, the gradle scripts are build with the `v` prefix in mind. I'm not sure if this change will break anything.

.ci/azure-pipelines.yml
```diff
@@ -78,15 +78,21 @@ jobs:
         env:
           BINTRAY_KEY: $(bintray_key)
 
+      - task: CopyFiles@2
+        displayName: 'Copy build assets'
+        inputs:
+          Contents: |
+            **/distributions/*-$(JELLYFIN_VERSION).zip
+            **/libs/*-$(JELLYFIN_VERSION).jar
+            **/libs/*-$(JELLYFIN_VERSION)-sources.jar
```
nielsvanvelzen, line 87: I just realized we don't include the platform-android .aar files. They are in the `./jellyfin-platform-android/build/outputs/aar/` directory and use `jellyfin-platform-android-debug.aar` and `jellyfin-platform-android-release.aar` as file names. We should to rename them (probably Gradle files) to use the version in the filename.

_Although this might be out of scope for this PR_

.ci/azure-pipelines.yml
h1dden-da3m0n: Morning, sorry it was kind of late when I threw this together. Ill have a look at the line in a few.

However, are you implying that the build artifacts/assets created during the `gradlew publish` are using `*-vX.Y.Z*` in their names? (which was the sole reason I hastily created this PR, since that would also be an explanation why the files where never found)

nielsvanvelzen: It should remove the `v` prefix for at least some parts the version is used but I'm not sure if it does that everywhere.

.ci/azure-pipelines.yml
h1dden-da3m0n: I have to sadly admit, that my skills in gradle scripting are lacking, so I don't feel comfortable editing those.
However, I can add a content search mask for the aar files within this PR if you want me to. :wink:

.ci/azure-pipelines.yml
h1dden-da3m0n: > `Build.SourceBranch` implies it's the name of a branch. For the apiclient (and Android repositories) we use tags to set the version and the branch can be anything (most of the time release branches though).

Okay just had a quick look at the documentation for this, and seems to be sound logic.
The [documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables-devops-services) reads as follows: 
> Build.SourceBranch:
> The branch of the triggering repo the build was queued for. Some examples: 
> * .....
>* **When your pipeline is triggered by a tag: refs/tags/your-tag-name** 
> * ....

h1dden-da3m0n: remains the question: do the artifacts that get created by the gradlew task account for the `vX.Y.Z` version in their name or just `X.Y.Z`? (from my experience with Maven Java artifacts only have a X.Y.Z version in their name without the `v`)

.ci/azure-pipelines.yml
```diff
@@ -78,15 +78,22 @@ jobs:
...
+            **/outputs/aar/*.arr
```
Maxr1998, line 88: I think this should be

```suggestion
            **/outputs/aar/*.aar
```
:wink:

Maxr1998 (changes requested):

.ci/azure-pipelines.yml
h1dden-da3m0n: yes :facepalm: :grin:

.ci/azure-pipelines.yml
h1dden-da3m0n: Okay, this may just be a shot in the dark, but with this code like this in the build.gradle.kts is it save to assume that the artifacts have no `v` in front of the version in their name:
```kts
// Versioning
allprojects {
	group = "org.jellyfin.apiclient"
	version = getProperty("jellyfin.version")?.removePrefix("v") ?: "SNAPSHOT"
}
```

Maxr1998 (commented):
To me, these changes look good, but I'm not very experienced with Azure pipelines, so I'm also waiting for a review from Niels.

h1dden-da3m0n:
Actually, pls hold this one for a moment. I need to read up on something mentioned [here](https://github.com/jellyfin/jellyfin-vue/pull/179#discussion_r503536524) (which also triggered me thinking about file names and that variable, aka. the source of this PR)

nielsvanvelzen:
I marked the PR as draft so we can't accidentally merge it.

h1dden-da3m0n:
Okay, I now read up on the syntax EraYaN recommended in the Vue PR I linked.
It really is a LOT simpler to read and understand IMHO.
I now replaces the Variable set task with a Job scoped variable, now all that is left is to hope this whole thing works with the next release :wink: 

**Relevant Documentation**
* [Set variables by using expressions](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#set-variables-by-using-expressions)
* [The Replace Expression](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#replace)

.ci/azure-pipelines.yml
nielsvanvelzen: I looked into it and it should be fine 👍

Maxr1998 (approved):

PR #128, opened 2020-10-14
Add openapi.json to git lfs

nielsvanvelzen:
This PR moves the openapi.json file to git-lfs.

Since this file is currently 1.5mb and will be updated regularly the repository size will keep increasing, cloning the repo might become slower with time. To fix that the file is now in git-lfs which is made for this exact use-case.

Atlassian has a good page describing git-lfs: https://www.atlassian.com/git/tutorials/git-lfs


nielsvanvelzen:
Codacy apparently doesn't like git-lfs lol

Maxr1998 (approved):

thornbill (approved):

PR #134, opened 2020-10-23
Add recommended server discovery

### LINKED ISSUE #101: Add function to determine best server from a list of candidates
> a function that takes the list of candidate server addresses and returns which one is valid
> - @thornbill 

It would be nice to have a function that extends the functionality of the function that will "normalize" a server string and return a list of candidates by connecting to the servers in the list and returning the "best". The best server will always prefer https over http but otherwise returns the first in the list that responds to a public system info.

nielsvanvelzen:
to find the best server candidate from a given input.

Closes #101

### DIFF
--- a/jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerDiscovery.kt
+++ b/jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerDiscovery.kt
@@ -28,6 +28,7 @@ class RecommendedServerDiscovery(
 		val api = SystemApi(client)
 
 		val startTime = System.currentTimeMillis()
+		@Suppress("TooGenericExceptionCaught")
 		val info = try {
 			api.getPublicSystemInfo()
 		} catch (err: ConnectException) {
@@ -50,7 +51,7 @@ class RecommendedServerDiscovery(
 		var points = 0
 
 		// Security
-		if (result.address.startsWith("https:")) points += 3
+		if (result.address.startsWith("https://")) points += 3
 
 		// Speed
 		when {
@@ -62,10 +63,10 @@ class RecommendedServerDiscovery(
 		// Compatibility
 		val version = result.systemInfo?.version?.let(ServerVersion::fromString)
 		if (version != null) {
-			points += 3
+			if (version >= Jellyfin.apiVersion) points += 1
 
-			if (version < Jellyfin.apiVersion) points -= 2
-			if (version < Jellyfin.recommendedVersion) points -= 1
+			if (version == Jellyfin.recommendedVersion) points += 2
+			else if (version > Jellyfin.recommendedVersion) points += 1
 		}
 
 		// Minimum amount of points: 0
@@ -113,7 +114,7 @@ class RecommendedServerDiscovery(
 			}.collect()
 		}.filter {
 			// Use [minimumScore] to filter out bad score matches
-			it.score.i >= minimumScore.i
+			it.score.score >= minimumScore.score
 		}
 	}
 }
--- a/jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerDiscovery.kt
+++ b/jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerDiscovery.kt
@@ -1,0 +1,119 @@
+package org.jellyfin.apiclient.discovery
+
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.withContext
+import org.jellyfin.apiclient.Jellyfin
+import org.jellyfin.apiclient.api.operations.SystemApi
+import org.jellyfin.apiclient.model.api.PublicSystemInfo
+import org.jellyfin.apiclient.model.discovery.ServerVersion
+import org.slf4j.LoggerFactory
+import java.net.ConnectException
+
+class RecommendedServerDiscovery(
+	private val jellyfin: Jellyfin
+) {
+	private val logger = LoggerFactory.getLogger("RecommendedServerDiscovery")
+
+	private data class SystemInfoResult(
+		val address: String,
+		val systemInfo: PublicSystemInfo?,
+		val responseTime: Long
+	)
+
+	private suspend fun getSystemInfoResult(address: String): SystemInfoResult? {
+		logger.info("Requesting public system info for $address")
+
+		val client = jellyfin.createApi(baseUrl = address)
+		val api = SystemApi(client)
+
+		val startTime = System.currentTimeMillis()
+		val info = try {
+			api.getPublicSystemInfo()
+		} catch (err: ConnectException) {
+			logger.debug("Could not connect to $address")
+			null
+		} catch (err: Exception) {
+			logger.error("Could not retrieve public system info for $address", err)
+			null
+		}
+		val endTime = System.currentTimeMillis()
+
+		return SystemInfoResult(
+			address = address,
+			systemInfo = if (info != null && info.status == 200) info.content else null,
+			responseTime = endTime - startTime,
+		)
+	}
+
+	private fun assignScore(result: SystemInfoResult, parentResult: SystemInfoResult? = null): RecommendedServerInfo {
+		var points = 0
+
+		// Security
+		if (result.address.startsWith("https:")) points += 3
+
+		// Speed
+		when {
+			result.responseTime < 500 -> points += 3
+			result.responseTime < 1500 -> points += 2
+			result.responseTime < 5000 -> points += 1
+		}
+
+		// Compatibility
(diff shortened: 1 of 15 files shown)


thornbill:
The build, the build is FAILING.

this is sewius

nielsvanvelzen:
image

jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerInfo.kt
```diff
@@ -0,0 +1,16 @@
...
+
+data class RecommendedServerInfo(
+	val address: String,
+	val responseTime: Long,
+	val score: RecommendedServerInfoScore,
+	val systemInfo: PublicSystemInfo?,
+	val parent: String?
+) {
+	/**
+	 * True when this server was not a part of the inputted addresses, false otherwise.
```
Maxr1998, line 13: Interestingly, "inputted" is a valid past participle of "input", but I think I'd prefer "entered" here regardless :sweat_smile:

jellyfin-model/src/main/kotlin/org/jellyfin/apiclient/model/discovery/ServerVersion.kt
```diff
@@ -0,0 +1,46 @@
...
+			{ it.minor },
+			{ it.patch }
+		)
+
+		fun fromString(str: String): ServerVersion? {
+			try {
+				val parts = str
+					.split('.') // Split into major, minor and patch
+					.also { require(it.size == 3) } // Check if we found 3 parts
+					.take(3) // Take first 3 values in case the input is bad
```
Maxr1998, line 30: Is the `take` even needed with the require before?

jellyfin-model/src/main/kotlin/org/jellyfin/apiclient/model/discovery/ServerVersion.kt
```diff
@@ -0,0 +1,46 @@
...
+				// Return server version
+				return ServerVersion(
+					major = parts[0],
+					minor = parts[1],
+					patch = parts[2]
+				)
+			} catch (err: IllegalArgumentException) {
+				// Invalid string
+				return null
+			}
```
Maxr1998, line 43: I think instead of a try catch there should be proper if/return statements..

samples/kotlin-console/src/main/kotlin/org/jellyfin/sample/console/cli/Ping.kt
```diff
@@ -11,8 +14,14 @@ class Ping(
 	private val jellyfin: Jellyfin
 ) : Subcommand("ping", "Pings a given server and retrieve basic system information") {
 	private val server by option(ArgType.String, description = "Url of the server", shortName = "s").required()
+	private val extended by option(ArgType.Boolean, description = "Normalize the URL and request", shortName = "e").default(false)
```
Maxr1998, line 17: The description isn't that obvious in what this actually does, but since it's only a sample it doesn't really matter I guess.

jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerDiscovery.kt
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		// Compatibility
+		val version = result.systemInfo?.version?.let(ServerVersion::fromString)
+		if (version != null) {
+			points += 3
+
+			if (version < Jellyfin.apiVersion) points -= 2
+			if (version < Jellyfin.recommendedVersion) points -= 1
+		}
```
Maxr1998, line 69: Wouldn't this also work the other way around? Also, could you use a `when` instead?

jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerInfoScore.kt
```diff
@@ -0,0 +1,9 @@
+package org.jellyfin.apiclient.discovery
+
+enum class RecommendedServerInfoScore(
+	val i: Int
```
Maxr1998, line 4: Even if repeated, it's way more insightful that way. Inline classes would be the perfect match here, but I don't know whether they're still marked as "unstable" or not..
```suggestion
	val score: Int
```

Maxr1998 (changes requested):
Sorry for taking so long, I (again) forgot about this. You should maybe bug me more often on Matrix :laughing:

jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerDiscovery.kt
```diff
@@ -0,0 +1,119 @@
...
+			systemInfo = if (info != null && info.status == 200) info.content else null,
+			responseTime = endTime - startTime,
+		)
+	}
+
+	private fun assignScore(result: SystemInfoResult, parentResult: SystemInfoResult? = null): RecommendedServerInfo {
+		var points = 0
+
+		// Security
+		if (result.address.startsWith("https:")) points += 3
```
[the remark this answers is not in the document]
nielsvanvelzen: Probably is but it is more readable like this IMO

jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerInfo.kt
nielsvanvelzen: According to the dictionary "input" means "something that is put in" and "enter" means "to come or go in". Since the value is already there the correct word would be "input".

jellyfin-model/src/main/kotlin/org/jellyfin/apiclient/model/discovery/ServerVersion.kt
nielsvanvelzen: Nop this was a leftover from the previous implementation.

samples/kotlin-console/src/main/kotlin/org/jellyfin/sample/console/cli/Ping.kt
nielsvanvelzen: I think I forgot to change it when I copied another boolean from a different file

jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerDiscovery.kt
nielsvanvelzen: Can't use a `when` since both if-statements can be effective making the score `-3`. Inverting the logic might be possible.

jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerInfoScore.kt
nielsvanvelzen: Forgot to change it

jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerDiscovery.kt
Maxr1998: Well it depends, if `recommendedVersion` is always greater than `apiVersion`, you could simplify the statement and use a when. No need for two checks then as well.

jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerDiscovery.kt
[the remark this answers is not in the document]
Maxr1998: Ok, you could still append `//` though ^^

jellyfin-core/src/main/kotlin/org/jellyfin/apiclient/discovery/RecommendedServerDiscovery.kt
nielsvanvelzen: They might differ so I'll keep the if-statements. I did invert them though

Maxr1998 (dismissed):
LGTM, thanks for the changes!

nielsvanvelzen:
I'm not sure if the suppress statements work in Codacy

thornbill:
@nielsvanvelzen it does! :wink: https://github.com/jellyfin/jellyfin-androidtv/pull/578/files#diff-edde6b8fb0af83594fb91de2133af5ed1695ae4e062bbcd7065da6220f1d51daR66

nielsvanvelzen:
That's good to know. I've actually installed (and configured) the Detekt plugin in Android Studio so it highlights the Detekt warnings now. The plugin is a bit unstable/slow but it helps finding the right place to put the suppress statement.

thornbill (approved):

Issue #130, opened 2020-10-23
Enable Kotlin 1.4 API mode


nielsvanvelzen:
https://kotlinlang.org/docs/reference/whatsnew14.html#explicit-api-mode-for-library-authors

All public modules (jellyfin-xx, not samples/generator) should use `explicitApi()`.

nielsvanvelzen:
Moved back to v1.0.0 because all (publicly available) modules now use explicitApi!

PR #138, opened 2020-10-31
Update Kotlinpoet to 1.7.2

nielsvanvelzen:
Changes:

- KotlinPoet 1.6.0 -> 1.7.2
- Regenerated OpenAPI output - now Kotlin 1.4 Explicit API mode compatible
- Switched to `explicitApiWarning` mode for api/model - will switch to the strict mode (uses errors instead of warnings) when I fix the non-generated code to be compatible (See #130)


Maxr1998 (approved):
I looked at all files carefully, so LGTM.

PR #140, opened 2020-11-11
Improvements in the device name normalization code

CarlosOlivo:
Now it should work the same way on mobile (device name) as well as on desktop (unit tests).

It also allows the use of some symbols that are part of [US-ASCII](https://en.wikipedia.org/wiki/ASCII#Character_set) and duplicate whitespace are removed.

Context: https://github.com/jellyfin/jellyfin-android/issues/230#issuecomment-724978586

### DIFF
--- a/jellyfin-platform-android/src/main/kotlin/org/jellyfin/apiclient/interaction/AndroidDevice.kt
+++ b/jellyfin-platform-android/src/main/kotlin/org/jellyfin/apiclient/interaction/AndroidDevice.kt
@@ -9,13 +9,14 @@ import org.jellyfin.apiclient.model.DeviceInfo
 /**
  * Helper function used in [androidDevice] to normalize device names:
  *
- * - Removes special characters from the name.
+ * - Removes non-US-ASCII-printable characters from the name.
  * - Trims the whitespace at the start and end of the name.
+ * - Removes repetitive whitespace from the name.
  *
  * Returns a copy of the device with the normalized name.
  */
 fun DeviceInfo.normalize() = copy(
-	name = name.replace("[^\\w\\s]".toRegex(), "").trim()
+	name = name.replace("[^\\x20-\\x7E]".toRegex(), "").trim().replace("\\s{2,}".toRegex(), " ")
 )
 
 @SuppressLint("HardwareIds")
--- a/jellyfin-platform-android/src/test/kotlin/org/jellyfin/apiclient/interaction/AndroidDeviceTests.kt
+++ b/jellyfin-platform-android/src/test/kotlin/org/jellyfin/apiclient/interaction/AndroidDeviceTests.kt
@@ -6,12 +6,22 @@ import kotlin.test.assertEquals
 
 class AndroidDeviceTests {
 	@Test
-	fun `Removes special characters`() {
-		assertEquals("s S9256GB_m", DeviceInfo("","Ã‘Ã¡Å¾Ã Å•'s S9+256GB..._\\m/\"").normalize().name)
+	fun `Removes non-US-ASCII-printable characters`() {
+		assertEquals("""'s S9+256GB..._\m/""", DeviceInfo("","""Ã‘Ã¡Å¾Ã Å•'s S9+256GB..._\m/""").normalize().name)
 	}
 
 	@Test
-	fun `Keeps normal characters`() {
+	fun `Keeps normal US-ASCII letters & numbers`() {
 		assertEquals("My Awesome Device 123", DeviceInfo("","My Awesome Device 123").normalize().name)
+	}
+
+	@Test
+	fun `Keeps normal US-ASCII punctuation & symbols`() {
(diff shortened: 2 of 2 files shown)


jellyfin-platform-android/src/main/kotlin/org/jellyfin/apiclient/interaction/AndroidDevice.kt
```diff
@@ -9,13 +9,14 @@ import org.jellyfin.apiclient.model.DeviceInfo
...
- * - Removes special characters from the name.
+ * - Removes non US-ASCII printable characters from the name.
  * - Trims the whitespace at the start and end of the name.
+ * - Remove duplicate whitespace from the name.
  *
  * Returns a copy of the device with the normalized name.
  */
 fun DeviceInfo.normalize() = copy(
-	name = name.replace("[^\\w\\s]".toRegex(), "").trim()
+	name = name.replace("[^\\x20-\\x7E]".toRegex(), "").trim().replace("\\s{2,}".toRegex(), " ")
```
Maxr1998, line 19: Looks good, but maybe it'd make sense to split the lines and add some documentation to them, like so:
```suggestion
	name = name
	        .replace("[^\\x20-\\x7E]".toRegex(), "") // Remove non-ASCII-printable characters (only allow SPACE to ~)
	        .trim() // Remove surrounding whitespace
	        .replace("\\s{2,}".toRegex(), " ") // Replace repetitive whitespace
```

jellyfin-platform-android/src/main/kotlin/org/jellyfin/apiclient/interaction/AndroidDevice.kt
```diff
@@ -9,13 +9,14 @@ import org.jellyfin.apiclient.model.DeviceInfo
 /**
  * Helper function used in [androidDevice] to normalize device names:
  *
- * - Removes special characters from the name.
+ * - Removes non US-ASCII printable characters from the name.
```
Maxr1998, line 12: I believe there's no other ASCII, so the extra US seems redundant.

```suggestion
 * - Removes non-ASCII-printable characters from the name.
```

Maxr1998 (approved):
Looking good, small suggestions regarding documentation though.

jellyfin-platform-android/src/main/kotlin/org/jellyfin/apiclient/interaction/AndroidDevice.kt
CarlosOlivo: It is the [encouraged name](https://www.iana.org/assignments/character-sets/character-sets.xhtml#character-sets-1:~:text=The%20use%20of%20the%20name%20US%2DASCII,encouraged.) established by the IANA so that it is not confused with the multiple [variants and derivations](https://en.wikipedia.org/wiki/ASCII#Variants_and_derivations) that exist of the mislabelled ASCII, but I can change it to ASCII if you want

jellyfin-platform-android/src/main/kotlin/org/jellyfin/apiclient/interaction/AndroidDevice.kt
CarlosOlivo: Should I remove part of the comment or leave both comments? (comment block + comment in code)

https://github.com/jellyfin/jellyfin-apiclient-java/blob/bf9179cbf0a45461acb558ef0a9069c610e4bfb0/jellyfin-platform-android/src/main/kotlin/org/jellyfin/apiclient/interaction/AndroidDevice.kt#L12-L14

Maxr1998: Oh nevermind, should've read more carefully. It's fine then.

jellyfin-platform-android/src/main/kotlin/org/jellyfin/apiclient/interaction/AndroidDevice.kt
Maxr1998: Interesting! Didn't know, that's probably more factual then. I'd still add the additional hyphens though.

CarlosOlivo: > I'd still add the additional hyphens though.

So `Removes non-US-ASCII-printable characters from the name.`?

Maxr1998: Yup.

nielsvanvelzen:
I did some testing and came to the conclusion that this issue only happens in the mobile android client because the actual issue is the webui not being able to serialize the name. I therfore think we should remove the normalization from the apiclient and add it to the android app only (at the part we send it to the JS code).

I'll make a PR for the android app in a bit.

PR #145, opened 2020-11-15
Enable explicit API mode in jellyfin-model

nielsvanvelzen:
Enables explicit API mode in the jellyfin-model module. Also added some really small naming/comment/error message fixes.

Part of #130


Maxr1998 (approved):
I trust the code-gen :P

nielsvanvelzen:
The generated code was done in #138

PR #150, opened 2020-11-21
Support nullable parameters in path

nielsvanvelzen:
- Support `null` as value for a path parameter
  - According to server devs we need to remove it from the path
- Added one additional test for nullable parameters
- Updated all tests to use public visibility - otherwise the IDE will complain because of strictApiMode
- Formatted some tests to make Detekt happy


Maxr1998 (approved):
LGTM. Are the explicit API changes (`public` in tests) intended for this PR?

nielsvanvelzen:
Somehow Android Studio complained about the unit tests (although the Gradle builds did not) so I just made them public. Doesn't hurt to do so.

I didn't bother to do it in a separate PR because it's a very small change.

Maxr1998:
Ok, let's merge then!

PR #154, opened 2020-12-02
Add option to generate URLs with access token added

nielsvanvelzen:
This will help when we request a URL for an image to load it with Glide but the image endpoint is authenticated.

As always: first commit is the actual change, the second updates the generated code.


nielsvanvelzen:
If the OpenAPI spec starts using the `security` object for operations I'll make a separate PR to set the default value of the parameter based on the security policy. (need auth = true, no auth required = false)

Maxr1998 (approved):

PR #155, opened 2020-12-04
Set default value for requireAuthentication

nielsvanvelzen:
OpenAPI spec from server now includes security metadata for all operations so we can determine the default value for this parameter now. Tested it with the latest openapi spec and it worked as expected.

Will update the spec (again) later when 10.7.0-rc.1 is released.

### DIFF
--- a/openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
+++ b/openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
@@ -75,8 +75,8 @@ class OpenApiApiServicesBuilder(
 					println("Missing return-type for operation $operationName (status-codes: ${operation.responses.keys})")
 
 				val requireAuthentication = operation.security
-					?.mapNotNull { it[Security.SECURITY_SCHEME] }
-					?.firstOrNull()
+					?.firstOrNull { requirement -> requirement.containsKey(Security.SECURITY_SCHEME) }
+					?.get(Security.SECURITY_SCHEME)
 					?.any(Security.AUTHENTICATION_POLICIES::contains)
 					?: false
 
--- a/openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
+++ b/openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
@@ -8,6 +8,7 @@ import net.pearx.kasechange.toCamelCase
 import org.jellyfin.openapi.builder.Builder
 import org.jellyfin.openapi.builder.api.ApiNameBuilder
 import org.jellyfin.openapi.constants.MimeType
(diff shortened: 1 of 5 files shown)


openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
```diff
@@ -73,12 +74,19 @@ class OpenApiApiServicesBuilder(
 				if (returnType == Unit::class.asTypeName() && "200" in operation.responses)
 					println("Missing return-type for operation $operationName (status-codes: ${operation.responses.keys})")
 
+				val requireAuthentication = operation.security
+					?.mapNotNull { it[Security.SECURITY_SCHEME] }
+					?.firstOrNull()
+					?.any(Security.AUTHENTICATION_POLICIES::contains)
+					?: false
```
nielsvanvelzen, line 81: I don't really like this but unfortunately the data structure is awful. It's a `List<Map<String, List<String>>>`. Feel free to suggest a better approach.

thornbill (dismissed):

openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
Maxr1998: What about that?
```kt
				val requireAuthentication = operation.security
					?.firstOrNull { requirement -> requirement.containsKey(Security.SECURITY_SCHEME) }
					?.get(Security.SECURITY_SCHEME)
					?.any { policy -> policy in Security.AUTHENTICATION_POLICIES }
					?: false
```

nielsvanvelzen: Ended up with 
```kt
val requireAuthentication = operation.security
	?.firstOrNull { requirement -> requirement.containsKey(Security.SECURITY_SCHEME) }
	?.get(Security.SECURITY_SCHEME)
	?.any(Security.AUTHENTICATION_POLICIES::contains)
	?: false
```

Maxr1998: I thought the `in` would be more readable, but it's fine this way.

Maxr1998 (approved):

openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
nielsvanvelzen: I'm quite the fan of function references haha

Maxr1998: Same! But in some cases the readability does suffer a bit. It's ok here, so I don't worry too much.

thornbill (approved):

PR #157, opened 2020-12-08
Fix lists incorrectly encoded in urls

nielsvanvelzen:
Before this fix a list would be encoded as `[]` for empty and `[1, 2, 3]` for filled. Now it will be ignored when empty and `1,2,3` when filled.

I've also added logging for requests.

### DIFF
--- a/jellyfin-api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
+++ b/jellyfin-api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
@@ -13,13 +13,14 @@ import io.ktor.util.*
 import kotlinx.serialization.json.Json
 import org.jellyfin.apiclient.model.ClientInfo
 import org.jellyfin.apiclient.model.DeviceInfo
+import org.slf4j.LoggerFactory
 import kotlin.collections.set
 
 public open class KtorClient(
 	override var baseUrl: String? = null,
 	public var accessToken: String? = null,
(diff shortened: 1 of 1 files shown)


jellyfin-api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
```diff
@@ -89,8 +90,17 @@ public open class KtorClient(
...
+					// Determine value
+					val value = when (rawValue) {
+						// Lists should be comma-separated
+						is Collection<*> -> if (rawValue.isNotEmpty()) rawValue.joinToString(",") else null
+						else -> rawValue.toString()
+					}
+
+					// Add not-null values
+					if (value != null)
+						parameters.append(key, value)
```
Maxr1998, line 103: Instead of the null-check, you could do `return@forEach` instead. Reads a bit neater imo.

nielsvanvelzen: I'd say that this is more readable. A return is easy to miss when reading the code while this indent clearly shows the branching.

Maxr1998: I'm not a fan of the extra null-check, but it's a matter of taste I guess ^^

Maxr1998 (approved):

thornbill (approved):

PR #159, opened 2020-12-08
Encode device name in authorization header

nielsvanvelzen:
Escape pretty much all characters to be safe.

References:

https://github.com/jellyfin/jellyfin/blob/79578521b8083588638ecdf70262346055b4db41/Emby.Server.Implementations/HttpServer/Security/AuthorizationContext.cs#L243-L284

https://github.com/jellyfin/jellyfin/blob/79578521b8083588638ecdf70262346055b4db41/Emby.Server.Implementations/HttpServer/Security/AuthorizationContext.cs#L286-L290

### DIFF
--- a/jellyfin-api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
+++ b/jellyfin-api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
@@ -98,12 +98,20 @@ public open class KtorClient(
 		}.buildString()
 	}
 
+	/**
+	 * This is a temporary workaround until we have a proper way to send device names to the server.
+	 * It will filter out all special characters.
+	 */
+	private fun String.encodeAuthorizationHeaderValue() = replace("[^\\w\\s]".toRegex(), "").trim()
+
 	override fun createAuthorizationHeader(): String? {
 		val params = mutableMapOf(
 			"client" to clientInfo.name,
 			"version" to clientInfo.version,
 			"deviceId" to deviceInfo.id,
-			"device" to deviceInfo.name
+			// Only encode the device name as it is user input
+			// other fields should be validated manually by the client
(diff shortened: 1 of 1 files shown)


jellyfin-api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
```diff
@@ -98,12 +98,20 @@ public open class KtorClient(
 		}.buildString()
 	}
 
+	/**
+	 * This is a temporary workaround until we have a proper way to send device names to the server.
+	 * It will filter out all special characters.
+	 */
+	private fun String.encodeAuthorizationHeaderValue() = replace("[^\\w\\s]".toRegex(), "").trim()
```
Maxr1998, line 105: Wouldn't URL-escaping be preferable here? Or doesn't the server like it then?

nielsvanvelzen: Server hates it :( It will just save it with urlencoding then

Maxr1998: I see, then we should keep it like this.

Maxr1998 (approved):

jellyfin-api/src/main/kotlin/org/jellyfin/apiclient/api/client/KtorClient.kt
thornbill, line 105: This should probably provide a default value in the case that all characters in the name are invalid.

nielsvanvelzen: image
No name works fine - it will just not display anything in the dashboard, left a device without a name and right my browser session with a name.

thornbill (approved):

PR #163, opened 2020-12-12
Move to GitHub Actions

nielsvanvelzen:
This PR:

- [x] Build & Test
- [x] Detekt lint results with problem matcher
- [x] Android lint results with problem matcher
- [x] Markdown lint results with problem matcher
- [x] Validate Gradle Wrapper
- [x] Automatically update Gradle Wrapper
- [x] Remove testing/building from Azure pipelines
- [x] Publish with GitHub Actions
- [x] Remove Azure 🎉 

Additional notes:

- We can remove Codacy from this repository as we now use linters via GitHub Actions.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,53 @@
+name: Lint
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Setup problem matcher
+        uses: xt0rted/markdownlint-problem-matcher@v1
+      - name: Run markdownlint
+        run: npx --package markdownlint-cli markdownlint '**/*.md'
+  code:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Setup Gradle cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Run detekt linter
+        run: ./gradlew --build-cache --no-daemon --info detekt
+        continue-on-error: true
+      - name: Report Detekt linter
+        uses: jwgmeligmeyling/checkstyle-github-action@v1.2
+        with:
+          name: Detekt
+          path: '**/detekt.xml'
+
+      - name: Run Android linter
+        run: ./gradlew --build-cache --no-daemon --info jellyfin-platform-android:lint
+        continue-on-error: true
+      - name: Report Android linter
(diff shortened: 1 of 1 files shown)


.github/workflows/gradlew-validate.yaml
```diff
@@ -0,0 +1,18 @@
+name: Gradle
```
h1dden-da3m0n, line 1: Not sure how GHA react to two workflows sharing one name? (+ IMHO a bit confusing :P )

Proposal rename:
* (this) Gradle Validate
* (gradle-update.yml) Gradle Update

.github/workflows/build.yaml
```diff
@@ -0,0 +1,34 @@
...
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      max-parallel: 2
+      matrix:
+        java: [8, 11, 15]
```
h1dden-da3m0n, line 16: May I ask why you now want to run on Java 8 and the NON LTS 15, when you ran every build before on Java 11

nielsvanvelzen: Good question. I wanted to include all LTS releases and the latest release. So that's why 15 is in the list.

Maxr1998: Java 8 is the default on Android platforms nowadays, while Java 15 is the latest stable version available. I think it makes sense.

.github/workflows/gradlew-validate.yaml
nielsvanvelzen: Ideally I wanted them in the same workflow file but that's impossible due too their different triggers. I chose to use the same name for both because GitHub always shows workflow and job name together.

h1dden-da3m0n: Okay, I see your use-case and if you say GH understands this then ignore my comment :wink:  :+1:

.github/workflows/build.yaml
h1dden-da3m0n: fair enough! thank you for the clarification :wink:

nielsvanvelzen: Note that the apiclient works with Java applications too, it's not specifically meant for Android apps.

h1dden-da3m0n (dismissed):

nielsvanvelzen:
I forgot to mention why this is still a draft while it appears to work: I don't like how the linter issues are reported right now. Both actions used (android/checkstyle) have a different method of reporting and I'm not a fan of that.
I started on my own action to fix this but ran into some limitations of the GitHub API so until I'll figure that out this PR is blocked.

When I eventually finish the action I'll update this PR and port it over to the android/androidtv repositories as I intended from the start :)

nielsvanvelzen:
Most of the features this PR originally added have now been merged in separate pull requests (including some more things!). The only missing part is linting. Unfortunately GitHub annotations still suck so I'm not sure how to add that.

From what I've read it's not possible to create more then 10 annotations using the actions toolkit without using the API. Using the API is not possible without a token though, which I don't want to add as that could provide access to the repository if someone creates a malicious pull request.

thornbill:
The limitation seems to be 10 per file which seems fine to me. If you have more than 10 you should probably check it locally lol.

nielsvanvelzen:
It's 10 for all results with the toolkit, it uses the API in this PR so that's why more show up

nielsvanvelzen:
You win @Maxr1998, created an issue (#220) to track my awesome linter idea.

PR #166, opened 2021-01-01
Update OpenAPI specification to 10.7~RC2

nielsvanvelzen:
Two commits, the first changes the code to detect binary files (changed in jellyfin/jellyfin#4762), the second updates the json+generated files.


Maxr1998 (approved):
I trust the codegen

PR #167, opened 2021-01-14
ci(deps): add dependabot config

h1dden-da3m0n:
### Description

While irregularly checking on dependabot I stumbled over the [announcement](https://github.blog/changelog/2020-12-18-dependabot-gradle-kotlin-and-composer-v2-support/) that it now supports `build.gradle.kts`.
I tested it very briefly, so I cannot guaranty that it would update all the dependencies, however to my eye it did so here the PR to add it.

Note: I took the liberty of creating this PR in all 3 repositories in the Jellyfin Org. that I know of using `build.gradle.kts`. I duplicated this text too.

### Changes

* add dependabot config

### Issues

* n/a

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10


nielsvanvelzen:
I'm quite sure this won't work since our dependencies are not easy to find. Instead of using a simple `implementation("dependencyspec")` we use variables from the buildSrc and even there they aren't just plaintext strings, often the version is a separate variable.

The only repository this change would work on is jellyfin-androidtv because that one isn't converted yet. Can you perhaps set the default branch of your fork to the the one of this PR so we can validate if it works?

h1dden-da3m0n:
> implementation("dependencyspec")

hmm I might have fired a bit too soon due to the success I seemingly had with the Android and AndroidTV repo.
I now see what u mean :thinking: . I see what u mean with the dependencies actually defined within `buildSrc/src/main/kotlin/Dependencies.kt `, and seems like my assumption of it working where just that: assumptions.

Sorry for any inconvenience I may have cause.

Closing since Dependabot is not capable of detecting these "advanced" dependency definitions.

nielsvanvelzen:
No problem, I hope that depandabot will be smart enough to support this though :p

PR #169, opened 2021-01-24
Update OpenAPI specification to 10.7~RC3

nielsvanvelzen:
- Change behavior for deprecated parameters in API operations to generate a secondary function

  The @deprecated annotation is not supported by Kotlin (although IntelliJ worked fine with it!). The new behavior will generate one function with the deprecated members removed and another one with them added. The latter will have a "Deprecated" suffix in the name and a @deprecated annotation on the function.
- Update OpenAPI specification to 10.7~RC3

The changes between RC2-RC3 are small so you can see the new behavior perfectly fine. I didn't notice that the old approach didn't work because this is the first release that adds deprecated parameters.

### DIFF
--- a/jellyfin-api/src/main/kotlin-generated/org/jellyfin/apiclient/api/operations/MediaInfoApi.kt
+++ b/jellyfin-api/src/main/kotlin-generated/org/jellyfin/apiclient/api/operations/MediaInfoApi.kt
@@ -9,6 +9,7 @@ import io.ktor.utils.io.ByteReadChannel
 import java.util.UUID
 import kotlin.Any
 import kotlin.Boolean
+import kotlin.Deprecated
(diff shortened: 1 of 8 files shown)


Maxr1998 (approved):
LGTM, but is the `Deprecated` suffix actually necessary? It's marked as `@Deprecated` anyway, so I don't see the need for an extra suffix..

nielsvanvelzen:
Right now we could use the same name as the parameters are different. But I explicitly used a different name for a few reasons:
- To make sure we can never get conflicting functions in case a parameter is replaced with a new one with a different type
- To force the app developer to make changes by throwing compile errors

Maxr1998:
Sounds good. Let's merge this then!

PR #170, opened 2021-01-24
Add service name hook

nielsvanvelzen:
New hook "ServiceNamehook":
  - Allows changing the service names for operations
  - Operations can now have multiple service names

New hook implementation "PlayStateServiceNameHook":
  - Renames the service name "PlaystateApi" to "PlayStateApi"
  - _This was the reason I made this pull request_

Constants:
  - Added new constant for the Api prefix

Naming:
  - Fixed Video3DFormat.kt -> Video3dFormat.kt, git on windows can be a pain with case sensitivity, had to do it partially via the GitHub UI.

I'm thinking about a new implementation of this hook that will add some operations to different services. Like adding the UniversalAudioApi operations to the AudioApi.


openapi-generator/src/main/kotlin/org/jellyfin/openapi/hooks/api/PlayStateServiceNameHook.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.jellyfin.openapi.constants.Strings
+import org.jellyfin.openapi.hooks.ServiceNameHook
+
+class PlayStateServiceNameHook : ServiceNameHook {
+	companion object {
+		const val PLAYSTATE_BAD = "Playstate${Strings.API_SERVICE_SUFFIX}"
+		const val PLAYSTATE_GOOD = "PlayState${Strings.API_SERVICE_SUFFIX}"
+	}
+
+	override fun mapServiceNames(operation: Operation, serviceNames: Set<String>) =
```
Maxr1998, line 13:
```suggestion
	override fun mapServiceNames(operation: Operation, serviceNames: Set<String>): Set<String> =
```

Maxr1998 (changes requested):

openapi-generator/src/main/kotlin/org/jellyfin/openapi/hooks/api/PlayStateServiceNameHook.kt
nielsvanvelzen: Generator doesn't use explicit API mode so it's not really necessary here. It inherits the return type anyway.

Maxr1998: Still, feells a bit cleaner imo and makes the return type more obvious. But whatever you prefer :P

nielsvanvelzen: I didn't do it for most code so it wouldn't make sense to only do it at this bit

Maxr1998 (approved):

PR #171, opened 2021-01-24
Add String.toUUID and String.toUUIDOrNull functions

nielsvanvelzen:
jellyfin-android needs to parse UUID's passed from the webclient. The `UUID.parseString` function in Java only allows hyphenated strings so we need to expose the apiclient UUID parser to parse the simple notation.

This PR adds two new extension functions to the String class to parse UUID's. The serializer now uses those functions to prevent duplicate code.

### DIFF
--- a/jellyfin-model/src/main/kotlin/org/jellyfin/apiclient/model/serializer/UUIDSerializer.kt
+++ b/jellyfin-model/src/main/kotlin/org/jellyfin/apiclient/model/serializer/UUIDSerializer.kt
@@ -8,6 +8,23 @@ import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
 import java.util.*
 
+private val UUID_REGEX = "^([a-z\\d]{8})([a-z\\d]{4})([a-z\\d]{4})([a-z\\d]{4})([a-z\\d]{12})\$".toRegex()
+
+/**
+ * Convert string to UUID. Accepts simple and hyphenated notations.
+ */
+public fun String.toUUID(): UUID = UUID.fromString(replace(UUID_REGEX, "$1-$2-$3-$4-$5"))
+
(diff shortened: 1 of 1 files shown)


jellyfin-model/src/main/kotlin/org/jellyfin/apiclient/model/serializer/UUIDSerializer.kt
```diff
@@ -8,23 +8,34 @@ import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
 import java.util.*
 
+private val UUID_REGEX = "^([a-z\\d]{8})([a-z\\d]{4})([a-z\\d]{4})([a-z\\d]{4})([a-z\\d]{12})\$".toRegex()
+
+/**
+ * Convert string to UUID. Accepts simple and hyphenated notations.
```
Maxr1998, line 14: Specify that this may throw on illegal UUID values

jellyfin-model/src/main/kotlin/org/jellyfin/apiclient/model/serializer/UUIDSerializer.kt
```diff
@@ -8,23 +8,34 @@ import kotlinx.serialization.encoding.Decoder
...
+ */
+public fun String.toUUID(): UUID = UUID.fromString(replace(UUID_REGEX, "$1-$2-$3-$4-$5"))
+
+/**
+ * Convert string to UUID or null if the string is not a uuid.
```
Maxr1998, line 19:
```suggestion
 * Convert string to UUID or null if the string is not an UUID.
```

Maxr1998 (changes requested):

Maxr1998 (approved):

PR #172, opened 2021-01-24
Support OpenAPI composition

nielsvanvelzen:
Starting from 10.7 RC4 / 10.7.1 / 10.8 the OpenAPI spec emits nullability by using composition to set the reference. This PR adds support for the composition and removes the NullableReferencesHook hack.

Draft until next server release.


Ladette19 (approved):

nielsvanvelzen:
Marking this as ready now, will regenerate sources when the RC4 spec is available.

Maxr1998 (approved):
LGTM. Should I merge this now or wait for updated sources?

nielsvanvelzen:
Merge is fine, I don't expect new changes before rc4

PR #180, opened 2021-02-20
Generate apiVersion and serverVersion constants

### LINKED ISSUE #161: Use OpenAPI document to set Jellyfin version in Jellyfin class
The generator should read `x-jellyfin-version` from the openapi document and put it somewhere in the code so we always know the "api version".

nielsvanvelzen:
- Generate apiVersion and serverVersion constants
- Update generated sources
- Use ApiConstants in Jellyfin.apiVersion

Closes #161


openapi-generator/src/main/kotlin/org/jellyfin/openapi/Generator.kt
```diff
@@ -35,6 +35,9 @@ class Generator(
 		.map(apiBuilder::build)
 		.map(fileSpecBuilder::build)
 
+	private fun createApiConstants(info: io.swagger.v3.oas.models.info.Info): FileSpec = openApiConstantsBuilder.build(info)
```
Maxr1998, line 38: I suppose an import would be reasonable here. Could use an alias import is `Info` isn't unique enough.

nielsvanvelzen: Android Studio somehow did it, fixed now.

Maxr1998 (approved):

PR #186, opened 2021-03-03
Use OperationBuilder as parent class in OperationUrlBuilder and remove duplicate code

nielsvanvelzen:
Fixes a different implementation for the buildParameter function.

The Url functions now support default values too :)


Maxr1998 (approved):
LGTM 👍

PR #189, opened 2021-03-10
Publish to Sonatype OSS

nielsvanvelzen:
- Remove Bintray publishing target
- Add Sonatype publishing target
- Change some other small things in build scripts

A future PR will update the CI (and moves it to GitHub).


buildSrc/src/main/kotlin/Pom.kt
```diff
@@ -0,0 +1,28 @@
+import org.gradle.api.publish.maven.MavenPublication
+
+fun MavenPublication.defaultPom() = pom {
```
Maxr1998, line 3: Couldn't this also be inlined? Since it's only used once and the content is quite specific, it's probably not necessary to pull this into a separate method.

build.gradle.kts
```diff
@@ -1,38 +1,61 @@
...
+}
 
-	// Publishing
-	plugins.apply("maven-publish")
-	publishing.repositories.jellyfinBintray(this)
+subprojects {
+	// Enable required plugins
+	apply<MavenPublishPlugin>()
+	apply<SigningPlugin>()
+	apply<io.gitlab.arturbosch.detekt.DetektPlugin>()
```
Maxr1998, line 34: Haven't used `subprojects` yet, but I suppose a `plugins` block wouldn't work?

build.gradle.kts
```diff
@@ -1,38 +1,61 @@
...
+			val signingKey = project.properties["signing.key"]?.toString()
+			val signingPassword = project.properties["signing.password"]?.toString()
+
+			useInMemoryPgpKeys(signingKey, signingPassword)
+			val publishing: PublishingExtension by project
+			sign(publishing.publications)
+		}
+
+		// Add POM to projects that use publishing
+		configure<PublishingExtension> {
```
Maxr1998, line 52: You should be able to use a `publishing` block instead, reads a bit nicer. Same for `signing`. [This](https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:deferred_configuration) might also help.

Maxr1998 (commented):
No full review yet, only quickly glanced over it. Looks mostly good, just a few comments.

buildSrc/src/main/kotlin/Pom.kt
nielsvanvelzen: It's for readability, the pom is quite large so I'd rather have it in a separate file so the build.gradle.kts files are easier to read.

build.gradle.kts
nielsvanvelzen: Nope, in `subprojects` and `allprojects` you need to use `plugins.apply("name")` or `apply<cls>()`. I used the second because type safety.

build.gradle.kts
nielsvanvelzen: Using `publishing {}` directly actually caused issues in other files, I still don't exactly know why. I've started to like the `configure<>` syntax though :p

build.gradle.kts
```diff
@@ -1,38 +1,63 @@
...
+}
 
-	// Publishing
-	plugins.apply("maven-publish")
-	publishing.repositories.jellyfinBintray(this)
+subprojects {
+	// Enable required plugins
+	apply<MavenPublishPlugin>()
+	apply<SigningPlugin>()
+	apply<io.gitlab.arturbosch.detekt.DetektPlugin>()
```
Maxr1998, line 34: Could you do an import for detekt?

nielsvanvelzen: I don't like imports in my build files 😅 (except for buildSrc)

Maxr1998: Ok then ^^

Maxr1998 (approved):

PR #190, opened 2021-03-11
Move build and testing CI to GitHub Actions

nielsvanvelzen:
- Add new workflow "apiclient-test"
  - this runs the test task which will build and test all modules
- Remove test and build jobs from azure
  - Also changed the triggers to only run on tags now so the azure file is correct
  - 1.0.0 will not use azure for publishing though

The CI file is largely copied from #163 but with changes in the caching keys

### DIFF
--- a/.github/workflows/apiclient-test.yaml
+++ b/.github/workflows/apiclient-test.yaml
@@ -1,0 +1,35 @@
+name: Apiclient Test
+
+on:
+  push:
+    branches:
+      - master
+      - release-*
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [8, 11, 15]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Setup Gradle cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-java-${{ matrix.java }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-java-${{ matrix.java }}-gradle-
+            ${{ runner.os }}-java-
+      - name: Run test task
+        run: ./gradlew --build-cache --no-daemon --info test
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -3,10 +3,6 @@ variables:
 
 trigger:
   batch: true
-  branches:
-    include:
-      - master
-      - release*
   tags:
     include:
       - '*'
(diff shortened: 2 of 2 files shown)


nielsvanvelzen:
Will do the publishing task next, depends on #189 though

.github/workflows/apiclient-test.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - master
+      - release-*
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      max-parallel: 2
```
ferferga, line 15: Why?

ferferga (commented):
I would also add dependabot for GH Actions updates

Additionally, I would add an empty newline between each job step, makes it much clearer and easy to read.

.github/workflows/apiclient-test.yaml
nielsvanvelzen: Java is a beast and may need quite some resources, in my earlier testing in #163 I noticed that it sometimes went out of memory. Limiting the maximum amount of builds to 2 fixed it.

nielsvanvelzen:
> I would add an empty newline between each job step, makes it much clearer and easy to read.

All other workflows in this repo and android/androidtv don't do it, I'm also not really a fan of it, the indentation already shows where a new step begins.

.github/workflows/apiclient-test.yaml
ferferga: That's strange. According to their [docs](https://docs.github.com/es/actions/reference/workflow-syntax-for-github-actions#) each run in a matrix runs inside their own worker.

[Workers have 7 GB as well](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) so I wonder how you could run out of memory as well

If you take a look at my [Jellyfin](https://github.com/ferferga/jellyfin) repo, where I'm working on the GH Actions migration for server for a future PR, you can see that I have a job that has a matrix for all the OSes and architectures, which dispatches 15 workers for a single job. [Maximum concurrent jobs are 20](https://docs.github.com/es/actions/reference/usage-limits-billing-and-administration), so that falls near the limits. 

I guess that max-parallel option was made for these kinds of scenarios like mine in server: If I push too quick in a row, the dispatched jobs only start after others have finished first. Some people might want their workflows to start as soon as possible, instead of parallelization, so that settings is tremendously useful for them.

However, imo, your setup isn't as "intensive" as the one I'm replacing in server or web ones (it's just a build process, no additional jobs for linting/testing like we have in Vue) so I don't see real benefit in removing parallelism, specially if there's no real proof that what you experienced before was a one-time thing or it could be reproduced consistently with the full parallel workflow.

ferferga:
That's a personal preference, for me the workflows without it are difficult to follow. This one is small though so that's not a huge deal.

You're the repo maintainer and PR-maker, so you choose :). I was just advising you in case the workflow might grow in the future

.github/workflows/apiclient-test.yaml
nielsvanvelzen: It might have changes, I'm quite sure they did run parallel on the same machine, sharing resources, in the past. Thanks for your explanation though. I'll remove this line :)

ferferga (approved):

h1dden-da3m0n (approved):

.github/workflows/apiclient-test.yaml
```diff
@@ -0,0 +1,35 @@
...
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Setup Gradle cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-java-${{ matrix.java }}-gradle-${{ hashFiles('**/*.gradle*') }}
```
Maxr1998, line 30: GitHub recommends also [hashing the gradle wrapper properties](https://github.com/actions/cache/blob/main/examples.md#java---gradle).

.github/workflows/apiclient-test.yaml
```diff
@@ -0,0 +1,35 @@
...
+          restore-keys: |
+            ${{ runner.os }}-java-${{ matrix.java }}-gradle-
+            ${{ runner.os }}-java-
+      - name: Run check task
```
Maxr1998, line 34:
```suggestion
      - name: Run test task
```

Maxr1998 (approved):

.github/workflows/apiclient-test.yaml
nielsvanvelzen: Good one, we want to clear cache when gradle updates of course

.github/workflows/apiclient-test.yaml
nielsvanvelzen: I have no excuse for this

Maxr1998 (approved):

h1dden-da3m0n (approved):

PR #191, opened 2021-03-13
Move publishing CI to GitHub actions

nielsvanvelzen:
This PR adds the publish CI to GitHub Actions. It triggers for both tags and the master branch. The master branch will release "latest-SNAPSHOT" builds, tags release a build based on the git tag. It also updates the Azure CI to only publish the GitHub releases, my next PR will move that to GitHub too.

Additional changes:
  - Update POM (src: https://central.sonatype.org/pages/requirements.html)
  - Use getProperty util function so we can pass properties via environment variables
  - Add logging to getProperty function, useful for debugging (shown with `--info` parameter when invoking Gradle)
 - Default to emtpy string when signing password is null

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -7,11 +7,6 @@ trigger:
     include:
       - '*'
 
-pr:
-  branches:
-    include:
-      - '*'
-
 jobs:
 
   - job: Publish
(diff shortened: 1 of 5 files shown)


Maxr1998 (approved):
LGTM. Any special reason for some of the reordering?

nielsvanvelzen:
In the build.gradle.kts the re-ordering of the apply() calls was to be consistent with the configuration calls below. In the POM file it was to be consistent with the maven sonatype requirement page. Nothing major.

Maxr1998:
Good to know, merging then.

PR #199, opened 2021-03-23
Publish javadoc using Dokka

nielsvanvelzen:
The Dokka plugin will generate javadoc files and those are added to the publications. Unfortunately Dokka loves memory so I had to increase it in the gradle.properties.

Partial fix for https://github.com/jellyfin/jellyfin-sdk-kotlin/issues/176#issuecomment-805068411

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,7 @@
 plugins {
 	id("io.github.gradle-nexus.publish-plugin") version "1.0.0"
 	id("io.gitlab.arturbosch.detekt").version(Dependencies.detektVersion)
+	id("org.jetbrains.dokka") version "1.4.30"
 }
 
 // Versioning
@@ -32,6 +33,7 @@ subprojects {
 	apply<SigningPlugin>()
 	apply<MavenPublishPlugin>()
 	apply<io.gitlab.arturbosch.detekt.DetektPlugin>()
+	apply<org.jetbrains.dokka.gradle.DokkaPlugin>()
 
 	// Add dependency repositories
 	repositories.defaultRepositories()
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,7 +1,7 @@
 plugins {
-	id("io.github.gradle-nexus.publish-plugin") version "1.0.0"
+	id("io.github.gradle-nexus.publish-plugin").version(Dependencies.nexusPublishPluginVersion)
 	id("io.gitlab.arturbosch.detekt").version(Dependencies.detektVersion)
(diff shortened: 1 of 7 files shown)


build.gradle.kts
```diff
@@ -1,6 +1,7 @@
 plugins {
 	id("io.github.gradle-nexus.publish-plugin") version "1.0.0"
 	id("io.gitlab.arturbosch.detekt").version(Dependencies.detektVersion)
+	id("org.jetbrains.dokka") version "1.4.30"
```
Maxr1998, line 4: Is that tied to the Kotlin version? If yes, could you use a constant/reference here?

jellyfin-platform-android/build.gradle.kts
```diff
@@ -49,13 +49,20 @@ tasks.create<Jar>("sourcesArtifact") {
...
+}
+
 // Because of limitations in the android plugin
 // the publishing definition should be inside the "afterEvaluate" block
 afterEvaluate {
 	publishing.publications.create<MavenPublication>("default") {
 		from(components["release"])
 
 		artifact(tasks["sourcesArtifact"])
+		artifact(tasks["javadocJar"])
```
Maxr1998, line 65: `javadocJar` reference won't work here?

jellyfin-model/build.gradle.kts
```diff
@@ -23,8 +23,15 @@ val sourcesJar by tasks.creating(Jar::class) {
 	from(sourceSets.getByName("main").allSource)
 }
 
+val javadocJar by tasks.creating(Jar::class) {
+	dependsOn(tasks.getByName("dokkaJavadoc"))
+	archiveClassifier.set("javadoc")
+	from("$buildDir/dokka/javadoc")
+}
```
Maxr1998, line 30: Can you somehow de-duplicate this code fragment?

build.gradle.kts
nielsvanvelzen: No but it should be moved to `buildSrc`. Forgot that

jellyfin-platform-android/build.gradle.kts
nielsvanvelzen: Nope because of the `afterEvaluate` block, Android is special

jellyfin-model/build.gradle.kts
nielsvanvelzen: Not really, gradle does a lot of magic during build script compilation and moving this caused everything to fail.

Maxr1998 (approved):

PR #218, opened 2021-04-10
Use experimental centralized dependency versions and type-safe project dependencies

nielsvanvelzen:
Also removed kotlin-stdlib dependency for the two projects still using it.

Plugins and buildscript dependencies cannot use the dependency catalog right now, I'm not sure if that is intended. I kept the old method for that (using buildSrc).


nielsvanvelzen:
Ignore the required checks for this PR as I already changed the repository setting to require the java 16 test (#217).

Maxr1998 (approved):
LGTM

PR #228, opened 2021-04-17
Add linting in CI using Detekt with Sarif output

nielsvanvelzen:


nielsvanvelzen:
Should've pushed it to this repo to test instead. Opening a new PR.

PR #231, opened 2021-04-18
Fix, suppress or modify all Detekt linter issues

nielsvanvelzen:
Mostly formatting/style changes.

### DIFF
--- a/jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/LocalServerDiscovery.kt
+++ b/jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/LocalServerDiscovery.kt
@@ -21,7 +21,8 @@ import java.net.SocketTimeoutException
  * the maximum amount of servers has been retrieved.
  */
 public class LocalServerDiscovery(
-	private val discoveryBroadcastAddressesProvider: DiscoveryBroadcastAddressesProvider = JavaNetBroadcastAddressesProvider()
+	private val discoveryBroadcastAddressesProvider: DiscoveryBroadcastAddressesProvider =
+		JavaNetBroadcastAddressesProvider(),
 ) {
 	public companion object {
 		public const val DISCOVERY_MESSAGE: String = "who is JellyfinServer?"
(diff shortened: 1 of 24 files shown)


jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/LocalServerDiscovery.kt
```diff
@@ -108,7 +111,11 @@ public class LocalServerDiscovery(
 
 		// Try reading incoming messages but with a maximum
 		val foundServers = mutableSetOf<String>()
-		for (i in 0..maxServers) {
+
+		var i = 0
+		while (i < maxServers) {
+			i++
```
Maxr1998, line 117: If I'm not mistaken, this changes the behavior of the code.
Probably intended since it seems to have been broken before, but what's wrong with a `for (i in 0 until maxServers)` loop?

Alternatively, a `do {} while ()` loop could make sense here. I don't really like the suggested syntax xD

nielsvanvelzen: Changed because of "Unused variable `i`.". Not too big a fan of the while either.

nielsvanvelzen: I went back to the for loop and added a `@Suppress` annotation

Maxr1998: I see. Would a `repeat(n) {}` and `return@repeat` possibly work?

nielsvanvelzen: Code uses both break and continue statements

Maxr1998: K, just suppressing the warning is fine.

Maxr1998 (approved):
Can't really comment on the detekt config, but the changes themselves look good to me.

PR #232, opened 2021-04-26
Add deviceInfo and clientInfo getters in Jellyfin class

nielsvanvelzen:
- Add deviceInfo and clientInfo getters in Jellyfin class
- More/improved docs

### DIFF
--- a/jellyfin-core/src/main/kotlin/org/jellyfin/sdk/Jellyfin.kt
+++ b/jellyfin-core/src/main/kotlin/org/jellyfin/sdk/Jellyfin.kt
@@ -8,10 +8,23 @@ import org.jellyfin.sdk.model.DeviceInfo
 import org.jellyfin.sdk.model.ServerVersion
 
 public class Jellyfin(
-	private val options: JellyfinOptions
+	private val options: JellyfinOptions,
 ) {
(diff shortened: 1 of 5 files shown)


jellyfin-core/src/main/kotlin/org/jellyfin/sdk/Jellyfin.kt
```diff
@@ -20,12 +33,14 @@ public class Jellyfin(
...
+	 *
+	 * Throws an [IllegalStateException] when the client or device information is missing.
 	 */
 	public fun createApi(
 		baseUrl: String? = null,
 		accessToken: String? = null,
-		clientInfo: ClientInfo? = options.clientInfo,
-		deviceInfo: DeviceInfo? = options.deviceInfo
+		clientInfo: ClientInfo? = this.clientInfo,
+		deviceInfo: DeviceInfo? = this.deviceInfo,
```
Maxr1998, line 43: The `this` shouldn't be necessary I guess. Also, you could just leave this as-is, increases "performance" cause one less method call.

nielsvanvelzen: The `this` is required because the parameter has the same name. I'll change it back to using `options` though.

Maxr1998 (approved):

PR #233, opened 2021-04-28
Rename recommendedVersion to minimumVersion

nielsvanvelzen:
The recommended version should always be the latest server version (which is stored in apiVersion). The minimumVersion should be the latest server version without significant API changes and is manually updated.

Also added an additional check to the server discovery score system. Non-jellyfin servers now return a BAD score for obvious reasons.

### DIFF
--- a/jellyfin-core/api/jellyfin-core.api
+++ b/jellyfin-core/api/jellyfin-core.api
@@ -11,7 +11,7 @@ public final class org/jellyfin/sdk/Jellyfin {
 
 public final class org/jellyfin/sdk/Jellyfin$Companion {
 	public final fun getApiVersion ()Lorg/jellyfin/sdk/model/ServerVersion;
-	public final fun getRecommendedVersion ()Lorg/jellyfin/sdk/model/ServerVersion;
+	public final fun getMinimumVersion ()Lorg/jellyfin/sdk/model/ServerVersion;
 }
 
 public final class org/jellyfin/sdk/JellyfinOptions {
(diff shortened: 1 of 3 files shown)


jellyfin-core/api/jellyfin-core.api
```diff
@@ -11,7 +11,7 @@ public final class org/jellyfin/sdk/Jellyfin {
 
 public final class org/jellyfin/sdk/Jellyfin$Companion {
 	public final fun getApiVersion ()Lorg/jellyfin/sdk/model/ServerVersion;
-	public final fun getRecommendedVersion ()Lorg/jellyfin/sdk/model/ServerVersion;
+	public final fun getMinimumVersion ()Lorg/jellyfin/sdk/model/ServerVersion;
```
nielsvanvelzen, line 14: ^ one breaking change - since we're still in beta this is fine. Needs to be changed in the Android app though.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -66,13 +67,14 @@ public class RecommendedServerDiscovery(
...
 		if (version != null) {
 			if (version >= Jellyfin.apiVersion) points += 1
-
-			if (version == Jellyfin.recommendedVersion) points += 2
-			else if (version > Jellyfin.recommendedVersion) points += 1
+			if (version >= Jellyfin.minimumVersion) points += 1
 		}
 
+		val productName = result.systemInfo?.productName
+		if (productName != null && !productName.equals("Jellyfin Server", ignoreCase = true)) points = 0
```
Maxr1998, line 74: Constant for this?

Maxr1998 (approved):

PR #234, opened 2021-05-02
Fix path variable replacement

Maxr1998:
Previously, path variables were only replaced properly if they were enclosed within slashes.
Cases like e.g. `/{one}.{two}/` caused a missing path variable error for `one}.{two`.

This commit fixes those cases by iterating over the path manually,
detecting variable start/end manually, while still replacing duplicate slashes.

### DIFF
--- a/jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
+++ b/jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
@@ -5,7 +5,6 @@ import io.ktor.client.call.*
 import io.ktor.client.features.*
 import io.ktor.client.features.json.*
 import io.ktor.client.features.json.serializer.*
-import io.ktor.client.features.websocket.*
 import io.ktor.client.request.*
 import io.ktor.client.statement.*
 import io.ktor.http.*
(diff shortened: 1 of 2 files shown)


jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
```diff
@@ -44,29 +43,76 @@ public open class KtorClient(
...
+					}
+					lastEnd = i
+
+					// Get path variable key
+					val name = path.substring(lastStart, lastEnd)
+
+					// Check if key is set
+					if (!pathParameters.containsKey(name)) {
+						throw MissingPathVariableError(name, path)
+					}
```
nielsvanvelzen, line 90:
```suggestion
					if (!pathParameters.containsKey(name)) throw MissingPathVariableError(name, path)
```

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
```diff
@@ -44,29 +43,76 @@ public open class KtorClient(
 		}
 	}
 
-	override fun createPath(path: String, pathParameters: Map<String, Any?>): String = path
-		.split('/')
-		.filterNot { it.isEmpty() }
-		.mapNotNull { rawName ->
-			val name = rawName.removeSurrounding(prefix = "{", suffix = "}")
+	override fun createPath(path: String, pathParameters: Map<String, Any?>): String {
```
nielsvanvelzen, line 46:
```suggestion
	override fun createPath(pathTemplate: String, pathParameters: Map<String, Any?>): String {
```

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
Maxr1998: The line was a bit long that way, so I prefer putting it below.

nielsvanvelzen:
```suggestion
					if (!pathParameters.containsKey(name))
						throw MissingPathVariableError(name, path)
```

Maxr1998: It seems since it's still below the 120 char line limit, I can also put it in the same line.

nielsvanvelzen (approved):

PR #237, opened 2021-05-05
Update Android Gradle plugin to 4.2

nielsvanvelzen:
- Update Android Gradle plugin to 4.2
- Add Android linter to linting workflow

Some notes: 4.2 compiles to Java 8 by default so we can remove those settings, SARIF output is disabled by default so it should be enabled. The `lint` task does not run `detekt` by default.

### DIFF
--- a/.github/workflows/sdk-lint.yaml
+++ b/.github/workflows/sdk-lint.yaml
@@ -27,8 +27,8 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
           restore-keys: |
             ${{ runner.os }}-gradle-
-      - name: Run detekt task
-        run: ./gradlew --build-cache --no-daemon --info detekt
+      - name: Run detekt and lint tasks
+        run: ./gradlew --build-cache --no-daemon --info detekt lint
       - name: Upload SARIF files
         uses: github/codeql-action/upload-sarif@v1
         if: ${{ always() }}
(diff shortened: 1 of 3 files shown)


.github/workflows/sdk-lint.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
       - name: Run detekt task
```
nielsvanvelzen, line 30:
```suggestion
      - name: Run detekt and lint tasks
```

nielsvanvelzen: fixed it, thank you kind reviewer

jellyfin-platform-android/build.gradle.kts
```diff
@@ -16,9 +16,6 @@ android {
 
 	compileOptions {
 		isCoreLibraryDesugaringEnabled = true
-
-		sourceCompatibility = JavaVersion.VERSION_1_8
-		targetCompatibility = JavaVersion.VERSION_1_8
```
Maxr1998, line 21: Even if it's enabled by default, shouldn't those be kept anyway?

nielsvanvelzen: No because they default ti Java 8 now

Maxr1998: I know, but when thinking about it yesterday when I read it in the blogpost, I thought it'd still make sense to keep those for future proofing. Then again, it'll probably stay on 8 for quite a while.

Maxr1998 (approved):

PR #239, opened 2021-05-05
Sort generated operations alphabetically

nielsvanvelzen:
- Sort generated operations alphabetically
- Update generated sources

So when reviewing the last spec update (10.7.5) I noticed some operations moved around, this PR makes sure they are always sorted so it's easier to see actual changes in the future.

Since there is no changed .api file you can be confident nothing major changed.


Maxr1998 (approved):
Should be fine ^^

PR #243, opened 2021-05-07
Refactor discovery code with better rules

nielsvanvelzen:
# Changes
## Address Candidates

- Drop baseurl addition - too small of a user base to do this for. Each candidate adds to the amount of HTTP requests to make.
- Always prioritize candidates

## Recommendation discover

- Add new score type: GREAT
- Document all score types to summarize what they mean
- Remove `includeAppendedServers`: too complex, not worth the amount of additional requests
- Completely rewritten the scoring system. Doesn't use points anymore but strict rules instead.
  - Easier to maintain and see how it works
  - Checks if the server actually responded now
  - Uses the new GREAT type if all checks succeed
- Moved some code around, tweaked some calls

## Discovery Service

- Change the kdoc to reference the used functions instead 
  - To prevent outdated comments
- Remove `includeAppendedServers`
- Remove `getRecommendedServer`. Client should determine on it's own which server to use.

# </>

I'd love to have some feedback on the rules. Do they make sense? I think they do.

### DIFF
--- a/jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/AddressCandidateHelper.kt
+++ b/jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/AddressCandidateHelper.kt
@@ -21,12 +21,17 @@ public class AddressCandidateHelper(
 		public const val JF_HTTPS_PORT: Int = 8920
 
 		/**
-		 * Default base url for Jellyfin
+		 * HTTP protocol prefix: http://
 		 */
-		public const val JF_BASE_URL: String = "/jellyfin/"
+		public const val PROTOCOL_HTTP: String = "http://"
+
+		/**
+		 * HTTPS protocol prefix: https://
+		 */
+		public const val PROTOCOL_HTTPS: String = "https://"
 	}
 
-	private val candidates = mutableListOf<Url>()
+	private val candidates = mutableSetOf<Url>()
 	private val prioritizeComparator = Comparator<Url> { a, b -> b.score() - a.score() }
 	private val logger = LoggerFactory.getLogger("AddressCandidateHelper")
 
@@ -38,8 +43,8 @@ public class AddressCandidateHelper(
 			candidates.add(URLBuilder().apply {
 				// Ktor doesn't like urls not starting with a protocol
 				// so we default to a http prefix
-				if (!input.startsWith("http://") && !input.startsWith("https://"))
-					takeFrom("http://$input")
+				if (!input.startsWith(PROTOCOL_HTTP) && !input.startsWith(PROTOCOL_HTTPS))
+					takeFrom(PROTOCOL_HTTP + input)
 				else
 					takeFrom(input)
 			}.build())
@@ -87,35 +92,13 @@ public class AddressCandidateHelper(
 	}
 
 	/**
-	 * Add a base url ([JF_BASE_URL]) for each candidate with a root or no path.
-	 */
-	public fun addBaseUrlCandidates() {
-		logger.debug("Adding base url candidates")
-
-		candidates
-			.filter { it.encodedPath.isEmpty() || it.encodedPath == "/" }
-			.forEach {
-				candidates.add(it.copy(encodedPath = JF_BASE_URL))
-			}
-	}
-
-	/**
-	 * Execute the fixes for common mistakes:
-	 *
-	 *   - Protocol
-	 *   - Port
-	 *
-	 * Not included:
-	 *
-	 *   - BaseUrl
+	 * Applies all fixes to the [input].
 	 */
 	public fun addCommonCandidates() {
 		logger.debug("Adding common candidates")
 
 		addProtocolCandidates()
 		addPortCandidates()
-		// Disabled by default since it doesn't affect most users
-		// addBaseUrlCandidates()
 	}
 
 	@Suppress("MagicNumber")
@@ -143,21 +126,13 @@ public class AddressCandidateHelper(
 	}
 
 	/**
-	 * Reorder all candidates based on priority.
+	 * Returns all unique candidates sorted by priority.
 	 *
-	 *   - HTTPS above HTTP
-	 *   - Jellyfin ports above default ports
-	 */
-	public fun prioritize() {
-		logger.debug("Prioritizing candidates")
-		candidates.sortWith(prioritizeComparator)
-	}
-
-	/**
-	 * Get all deduplicated candidate URLs as strings.
-	 * Call [prioritize] before if a sorted list is desired.
+	 * The priority is based on a few rules:
+	 * - HTTPS before HTTP
+	 * - Jellyfin ports before protocol default ports
 	 */
 	public fun getCandidates(): List<String> = candidates
+		.sortedWith(prioritizeComparator)
 		.map { it.toString() }
-		.distinct()
 }
--- a/jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/DiscoveryService.kt
+++ b/jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/DiscoveryService.kt
@@ -1,8 +1,6 @@
 package org.jellyfin.sdk.discovery
 
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.collect
-import kotlinx.coroutines.flow.takeWhile
 import org.jellyfin.sdk.Jellyfin
 import org.jellyfin.sdk.model.api.ServerDiscoveryInfo
 
@@ -11,7 +9,7 @@ import org.jellyfin.sdk.model.api.ServerDiscoveryInfo
  */
 public class DiscoveryService(
 	private val jellyfin: Jellyfin,
-	private val discoveryBroadcastAddressesProvider: DiscoveryBroadcastAddressesProvider
+	private val discoveryBroadcastAddressesProvider: DiscoveryBroadcastAddressesProvider,
 ) {
 	private val localServerDiscovery by lazy {
 		LocalServerDiscovery(discoveryBroadcastAddressesProvider)
@@ -22,44 +20,34 @@ public class DiscoveryService(
 	}
 
 	/**
-	 * Parses the given [input] and tries to fix common mistakes.
+	 * Parses the given [input] and tries to fix common mistakes. The returned candidates are in order
+	 * of most secure followed by most likely to work.
 	 *
-	 *   - Add protocol (https and http)
-	 *   - Add port (8096 for http and 8920 for https)
-	 *   - Add baseurl (/jellyfin)
-	 *
-	 * The returned candidates are in order of most secure followed by most likely to work.
 	 * See [AddressCandidateHelper] for more information.
 	 */
 	public fun getAddressCandidates(input: String): List<String> = AddressCandidateHelper(input).run {
 		addCommonCandidates()
-		prioritize()
 		getCandidates()
 	}
 
 	/**
-	 * Connects to the [servers] and returns a list of responding servers ordered by input order.
-	 * Uses multiple rules to determine the recommendation like:
-	 *
-	 *   - HTTPS above HTTP
-	 *   - Higher Jellyfin version
-	 *   - Respond time
-	 *
-	 * Also adds the addresses reported by the server into the mix when visiting the public server
-	 * information. Reported addresses are inserted before their parent server.
+	 * Connects to the [servers] and returns a flow of scored servers in the same order as the input.
+	 * Uses the [RecommendedServerDiscovery] rules to determine the recommendation.
 	 *
 	 * If the input in [servers] is the output from [getAddressCandidates] the optimal server to use
-	 * will be the first returned server with a [RecommendedServerInfoScore] of GOOD.
+	 * will be the first returned server with a [RecommendedServerInfoScore.GREAT] score. If no server
+	 * with [RecommendedServerInfoScore.GREAT] is returned the best option is the first server with a
+	 * [RecommendedServerInfoScore.GOOD] score. The [RecommendedServerInfoScore.OK] and
+	 * [RecommendedServerInfoScore.BAD] scored are not recommended for connection.
 	 *
-	 * Optionally use [getRecommendedServer] to make this selection automatically.
+	 * Output is not sorted, a BAD match could be returned before a GOOD match.
+	 * **Never just use the first result**.
 	 */
 	public suspend fun getRecommendedServers(
 		servers: List<String>,
-		includeAppendedServers: Boolean = true,
-		minimumScore: RecommendedServerInfoScore = RecommendedServerInfoScore.BAD
+		minimumScore: RecommendedServerInfoScore = RecommendedServerInfoScore.BAD,
 	): Flow<RecommendedServerInfo> = recommendedServerDiscovery.discover(
 		servers = servers,
-		includeAppendedServers = includeAppendedServers,
 		minimumScore = minimumScore
 	)
 
@@ -68,36 +56,11 @@ public class DiscoveryService(
 	 */
 	public suspend fun getRecommendedServers(
 		input: String,
-		includeAppendedServers: Boolean = true,
-		minimumScore: RecommendedServerInfoScore = RecommendedServerInfoScore.BAD
+		minimumScore: RecommendedServerInfoScore = RecommendedServerInfoScore.BAD,
 	): Flow<RecommendedServerInfo> = getRecommendedServers(
 		servers = getAddressCandidates(input),
-		includeAppendedServers = includeAppendedServers,
 		minimumScore = minimumScore
 	)
-
-	/**
-	 * Utility function that calls [getRecommendedServers] for inputted [servers] and returns the
-	 * best candidate. Returns when a server with a score of [RecommendedServerInfoScore.GOOD] is
-	 * found or otherwise collects all servers and returns the best one based on order and score.
-	 */
-	public suspend fun getRecommendedServer(
-		servers: List<String>,
-		includeAppendedServers: Boolean = true
-	): RecommendedServerInfo? {
-		var best: RecommendedServerInfo? = null
-
-		getRecommendedServers(servers, includeAppendedServers).takeWhile {
-			// Select if it's better than current
-			if (best == null || it.score.score > best!!.score.score)
-				best = it
-
-			// Take while score is not GOOD (highest possible value)
-			best!!.score != RecommendedServerInfoScore.GOOD
-		}.collect()
-
-		return best
-	}
 
 	/**
 	 * Discover servers on the local network. Uses the [discoveryBroadcastAddressesProvider] to
@@ -105,7 +68,7 @@ public class DiscoveryService(
 	 */
 	public fun discoverLocalServers(
 		timeout: Int = LocalServerDiscovery.DISCOVERY_TIMEOUT,
-		maxServers: Int = LocalServerDiscovery.DISCOVERY_MAX_SERVERS
+		maxServers: Int = LocalServerDiscovery.DISCOVERY_MAX_SERVERS,
 	): Flow<ServerDiscoveryInfo> = localServerDiscovery.discover(
 		timeout = timeout,
 		maxServers = maxServers
--- a/jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
+++ b/jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
@@ -1,24 +1,31 @@
 package org.jellyfin.sdk.discovery
 
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.asFlow
+import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.withContext
 import org.jellyfin.sdk.Jellyfin
+import org.jellyfin.sdk.api.client.HttpClientOptions
+import org.jellyfin.sdk.api.client.Response
+import org.jellyfin.sdk.api.client.exception.InvalidContentException
+import org.jellyfin.sdk.api.client.exception.InvalidStatusException
+import org.jellyfin.sdk.api.client.exception.TimeoutException
 import org.jellyfin.sdk.api.operations.SystemApi
 import org.jellyfin.sdk.model.ServerVersion
 import org.jellyfin.sdk.model.api.PublicSystemInfo
 import org.slf4j.LoggerFactory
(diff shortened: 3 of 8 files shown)


nielsvanvelzen:
Additional changes (just pushed):

- Change discovery command in cli sample to show candidate + recommendation results
- Add option to KtorClient to disable redirection following and set the connection timeout
  - Uses the current behavior by default
- Change `content` in `Response<T>` to nullable
  - `getValue` is still not-null (what you get when using `val x by api.call()`)
  - If you know what you're doing you can use the new getter `requireContent`

nielsvanvelzen:
Using the changed discovery command in the CLI, backported to master branch. This is the difference right now:

**New discovery code**
```
Starting discovery for 192.168.1.10
Found 5 candidates
https://192.168.1.10: BAD (replied in 2216ms, system information not found)
https://192.168.1.10:8096: BAD (replied in 31ms, system information not found)
https://192.168.1.10:8920: BAD (replied in 2029ms, system information not found)
http://192.168.1.10: BAD (replied in 2022ms, system information not found)
http://192.168.1.10:8096: GREAT (replied in 49ms, system information found)
```

**Old discovery code**
```
Starting discovery for 192.168.1.10
Found 5 candidates
https://192.168.1.10: OK (replied in 2226ms, system information not found)
https://192.168.1.10:8096: GOOD (replied in 31ms, system information not found)
https://192.168.1.10:8920: OK (replied in 2020ms, system information not found)
http://192.168.1.10: BAD (replied in 2059ms, system information not found)
http://192.168.1.10:8096: OK (replied in 2ms, system information found)
http://192.168.1.10:8096: OK (replied in 49ms, system information found)
```

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/AddressCandidateHelper.kt
```diff
@@ -143,21 +126,14 @@ public class AddressCandidateHelper(
...
-	 * Get all deduplicated candidate URLs as strings.
-	 * Call [prioritize] before if a sorted list is desired.
+	 * The priority is based on a few rules:
+	 * - HTTPS before HTTP
+	 * - Jellyfin ports before protocol default ports
 	 */
 	public fun getCandidates(): List<String> = candidates
+		.sortedWith(prioritizeComparator)
 		.map { it.toString() }
 		.distinct()
```
Maxr1998, line 138: Probably not necessary anymore with the set? At least if the URL is bijective to its string representation wrt uniqueness/equality.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -27,103 +29,95 @@ public class RecommendedServerDiscovery(
...
+	}
+
+	private suspend fun getSystemInfoResult(address: String): SystemInfoResult {
 		logger.info("Requesting public system info for $address")
 
-		val client = jellyfin.createApi(baseUrl = address)
+		val client = jellyfin.createApi(
+			baseUrl = address,
+			httpClientOptions = HttpClientOptions(
+				followRedirects = false,
```
Maxr1998, line 69: Shouldn't the SDK instead use the redirected URL for scoring? Or do we not want to support that?

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -27,103 +29,95 @@ public class RecommendedServerDiscovery(
...
-		val info = try {
-			api.getPublicSystemInfo()
-		} catch (err: ConnectException) {
-			logger.debug("Could not connect to $address", err)
-			null
-		} catch (err: Exception) {
-			logger.error("Could not retrieve public system info for $address", err)
-			null
+		val info: Response<PublicSystemInfo>?
+		val responseTime = measureTimeMillis {
```
Maxr1998, line 76: Note for later: when `measureTimedValue` is not experimental anymore, that could be used instead.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -27,103 +29,95 @@ public class RecommendedServerDiscovery(
...
+	public suspend fun discover(
+		servers: Flow<String>,
+		minimumScore: RecommendedServerInfoScore,
+	): Flow<RecommendedServerInfo> = withContext(Dispatchers.IO) {
+		servers
+			.map(::getSystemInfoResult)
+			.mapNotNull(::assignScore)
+			.filter {
+				// Use [minimumScore] to filter out bad score matches
+				it.score.score >= minimumScore.score
```
Maxr1998, line 109:
```suggestion
			.filter { server ->
				// Use [minimumScore] to filter out bad score matches
				server.score.score >= minimumScore.score
```

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerInfoScore.kt
```diff
@@ -3,7 +3,26 @@ package org.jellyfin.sdk.discovery
 public enum class RecommendedServerInfoScore(
 	internal val score: Int
 ) {
```
Maxr1998, line 5: Should be `Comparable` maybe?

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
Maxr1998, line 109: If `RecommendedServerInfoScore` was comparable, this could be shorter.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/DiscoveryService.kt
```diff
@@ -22,44 +20,31 @@ public class DiscoveryService(
...
-	 *   - Higher Jellyfin version
-	 *   - Respond time
-	 *
-	 * Also adds the addresses reported by the server into the mix when visiting the public server
-	 * information. Reported addresses are inserted before their parent server.
+	 * Uses the [RecommendedServerDiscovery] rules to determine the recommendation.
 	 *
 	 * If the input in [servers] is the output from [getAddressCandidates] the optimal server to use
-	 * will be the first returned server with a [RecommendedServerInfoScore] of GOOD.
+	 * will be the first returned server with a [RecommendedServerInfoScore] of GREAT or GOOD.
```
Maxr1998, line 38: Does that still apply?
Might be a bit misunderstandable regardless. Maybe swap it and write:
"the first server with a score of GREAT or GOOD will be the optimal server to use".

Also, KDoc link for `GREAT` and `GOOD` please.

Maxr1998 (changes requested):

Maxr1998:
> * Add option to KtorClient to disable redirection following and set the connection timeout
>   
>   * Uses the current behavior by default
> * Change `content` in `Response<T>` to nullable
>   
>   * `getValue` is still not-null (what you get when using `val x by api.call()`)
>   * If you know what you're doing you can use the new getter `requireContent`
> * Don't follow redirects in recommended server discovery to prevent recommending a URL that will always redirect

Those should definitely be mentioned in the PR title and description. I would even argue that they should be split into a different PR, but that's probably impractical now.

nielsvanvelzen:
I did some git magic for that change, it's already the first commit in this PR now so I can cherrypick it to a separate PR 😉

Maxr1998:
> I did some git magic for that change, it's already the first commit in this PR now so I can cherrypick it to a separate PR 😉 

You should do that then. Also rewrite the description and include the nullability change in there.

nielsvanvelzen:
Temporary changing this PR to draft until #246 is merged and this one is rebased.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
nielsvanvelzen: True, although the calculation later only needs the milliseconds anyway

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/AddressCandidateHelper.kt
nielsvanvelzen: Correct, left-over from the list implementation

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
nielsvanvelzen: Ideally we follow redirects to change the inputted url. Someone could then make `jellyfin.myawesomewebsite.com` to something different. Could be useful for loadbalancing etc.
It's quite hard to do so without complicating the code too much so I didn't bother with it _for now_.

Maxr1998: I meant that you could include `info` with that as well :P
But yeah, it doesn't matter for now.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerInfoScore.kt
nielsvanvelzen: Can't override `compareTo` from an enum.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
Maxr1998: I meant that you could include `info` with that as well :P
But yeah, it doesn't matter for now.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
nielsvanvelzen: Can't override `compareTo` from an enum.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
Maxr1998: Fair enough.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerInfoScore.kt
Maxr1998: Oh, right. Should the enum order be enough then though?

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/DiscoveryService.kt
```diff
@@ -22,44 +20,34 @@ public class DiscoveryService(
...
 	 * See [AddressCandidateHelper] for more information.
 	 */
 	public fun getAddressCandidates(input: String): List<String> = AddressCandidateHelper(input).run {
 		addCommonCandidates()
-		prioritize()
 		getCandidates()
 	}
 
 	/**
 	 * Connects to the [servers] and returns a list of responding servers ordered by input order.
```
Maxr1998, line 34:
```suggestion
	 * Connects to the [servers] and returns a flow of responding servers according to the input order.
```

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/DiscoveryService.kt
```diff
@@ -22,44 +20,34 @@ public class DiscoveryService(
...
 	 * If the input in [servers] is the output from [getAddressCandidates] the optimal server to use
-	 * will be the first returned server with a [RecommendedServerInfoScore] of GOOD.
+	 * will be the first returned server with a [RecommendedServerInfoScore.GREAT] score. If no server
+	 * with [RecommendedServerInfoScore.GREAT] is returned the best option is the first server with a
+	 * [RecommendedServerInfoScore.GOOD] score. The [RecommendedServerInfoScore.OK] and
+	 * [RecommendedServerInfoScore.BAD] scored are not recommended for connection.
 	 *
-	 * Optionally use [getRecommendedServer] to make this selection automatically.
+	 * Output is not sorted, a BAD match could be returned before a GOOD match. **never just
+	 * use the first result**.
```
Maxr1998, line 44:
```suggestion
	 * Output is not sorted, a BAD match could be returned before a GOOD match.
	 * **Never just use the first result**.
```

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -27,103 +31,101 @@ public class RecommendedServerDiscovery(
...
+			// All other checks completed, server should have a great connection
+			else -> RecommendedServerInfoScore.GREAT
+		}
+
+		return RecommendedServerInfo(
+			result.address,
+			result.responseTime,
+			score,
+			result.systemInfo,
+			parentResult?.address
```
Maxr1998, line 61:
```suggestion
			parentResult?.address,
```

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -27,103 +31,101 @@ public class RecommendedServerDiscovery(
...
+
+	private suspend fun getSystemInfoResult(address: String): SystemInfoResult {
 		logger.info("Requesting public system info for $address")
 
-		val client = jellyfin.createApi(baseUrl = address)
+		val client = jellyfin.createApi(
+			baseUrl = address,
+			httpClientOptions = HttpClientOptions(
+				followRedirects = false,
+				timeout = HTTP_TIMEOUT
```
Maxr1998, line 72:
```suggestion
				timeout = HTTP_TIMEOUT,
```

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -27,103 +31,101 @@ public class RecommendedServerDiscovery(
...
+			)
```
Maxr1998, line 73:
```suggestion
			),
```

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -27,103 +31,101 @@ public class RecommendedServerDiscovery(
...
+	 * Discover all servers in the [servers] flow and retrieve the public system information to assign a score.
+	 * Returned servers are not necessarily in order of score. Use [minimumScore] to automatically remove bad matches.
+	 */
+	public suspend fun discover(
+		servers: Flow<String>,
+		minimumScore: RecommendedServerInfoScore,
+	): Flow<RecommendedServerInfo> = withContext(Dispatchers.IO) {
+		servers
+			.map(::getSystemInfoResult)
+			.mapNotNull(::assignScore)
```
Maxr1998, line 114: Shouldn't it be the other way around?

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -27,103 +31,101 @@ public class RecommendedServerDiscovery(
 		val responseTime: Long,
 	)
 
-	private suspend fun getSystemInfoResult(address: String): SystemInfoResult? {
+	private val logger = LoggerFactory.getLogger("RecommendedServerDiscovery")
+
+	@Suppress("MagicNumber")
+	private fun assignScore(result: SystemInfoResult, parentResult: SystemInfoResult? = null): RecommendedServerInfo {
```
Maxr1998, line 37: What's that parent meant for?

Maxr1998 (changes requested):

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
nielsvanvelzen: - Remove includeAppendedServers: too complex, not worth the amount of additional requests

if only I removed it everywhere...

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/DiscoveryService.kt
```diff
@@ -22,44 +20,34 @@ public class DiscoveryService(
...
 	 * If the input in [servers] is the output from [getAddressCandidates] the optimal server to use
-	 * will be the first returned server with a [RecommendedServerInfoScore] of GOOD.
+	 * will be the first returned server with a [RecommendedServerInfoScore.GREAT] score. If no server
+	 * with [RecommendedServerInfoScore.GREAT] is returned the best option is the first server with a
+	 * [RecommendedServerInfoScore.GOOD] score. The [RecommendedServerInfoScore.OK] and
+	 * [RecommendedServerInfoScore.BAD] scored are not recommended for connection.
 	 *
-	 * Optionally use [getRecommendedServer] to make this selection automatically.
+	 * Output is not sorted, a BAD match could be returned before a GOOD match.
+	 * **never just use the first result**.
```
Maxr1998, line 44:
```suggestion
	 * **Never just use the first result**.
```

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
Maxr1998: Ah ok, well then.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerInfoScore.kt
Maxr1998: That's a good point. A comment could help, but we should probably just keep it like this.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -27,103 +34,100 @@ public class RecommendedServerDiscovery(
...
-		val score = when {
-			points < 3 -> RecommendedServerInfoScore.BAD
-			points < 6 -> RecommendedServerInfoScore.OK
-			else -> RecommendedServerInfoScore.GOOD
-		}
-
-		return RecommendedServerInfo(result.address, result.responseTime, score, result.systemInfo, parentResult?.address)
+	/**
+	 * Discover all servers in the [servers] flow and retrieve the public system information to assign a score.
+	 * Returned servers are not necessarily in order of score. Use [minimumScore] to automatically remove bad matches.
```
Maxr1998, line 108:
```suggestion
	 * Returned servers are not ordered by score. Use [minimumScore] to automatically remove bad matches.
```

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/DiscoveryService.kt
```diff
@@ -22,44 +20,34 @@ public class DiscoveryService(
...
-	 * Uses multiple rules to determine the recommendation like:
-	 *
-	 *   - HTTPS above HTTP
-	 *   - Higher Jellyfin version
-	 *   - Respond time
-	 *
-	 * Also adds the addresses reported by the server into the mix when visiting the public server
-	 * information. Reported addresses are inserted before their parent server.
+	 * Connects to the [servers] and returns a flow of responding servers according to the input order.
+	 * Uses the [RecommendedServerDiscovery] rules to determine the recommendation.
```
Maxr1998, line 35: Actually, it's not necessarily responding ones.

```suggestion
	 * Connects to the [servers], rates them and returns the results as a flow, in the same order as the input.
	 * Uses the [RecommendedServerDiscovery] rules to determine the recommendation.
```

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -27,103 +34,100 @@ public class RecommendedServerDiscovery(
...
+		minimumScore: RecommendedServerInfoScore,
+	): Flow<RecommendedServerInfo> = withContext(Dispatchers.IO) {
+		servers
+			.map(::getSystemInfoResult)
+			.map(::assignScore)
+			.filter { serverInfo ->
+				// Use [minimumScore] to filter out bad score matches
+				serverInfo.score.score >= minimumScore.score
+			}
 	}
```
Maxr1998, line 121: Should use [flowOn](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/flow-on.html) (I don't even know if withContext works here).

nielsvanvelzen: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/flow-on.html

The withContext works, flowOn can be used to run a part of the flow in a different context.

Maxr1998: lmao, I'm blind. I linked the exact same page. Didn't see the `withContext` in the example.

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/DiscoveryService.kt
Maxr1998: I like your change even more :+1:

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
nielsvanvelzen: I'm blind too because I did not notice the link in your reply 😂

Maxr1998 (approved):

PR #245, opened 2021-05-07
Set nullability for query parameters based on the required attribute instead

nielsvanvelzen:
Starting from Jellyfin 10.8 the nullability is not set for query parameters. We should use the requires property instead. Apparently this also fixes some operations for our current 10.7 spec


nielsvanvelzen:
Imagine forgetting to `apiDump` ~~already closed Android Studio thought~~

Maxr1998 (approved):

PR #246, opened 2021-05-09
Add HttpClientOptions and make respone body nullable

nielsvanvelzen:
- Add option to KtorClient to disable redirection following and set the connection timeout
  - Uses the current behavior by default
  - Timeout is now 6 seconds by default (was 10)
- Change `content` in `Response<T>` to nullable
  - `getValue` is still not-null (what you get when using `val x by api.call()`)
  - If you know what you're doing you can use the new getter `requireContent`


nielsvanvelzen:
Small note:

If the SDK is used like recommended in the docs (using the `by` syntax) no changes should be made in clients. Only advanced usage where the response class is used might need updating.

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/Response.kt
```diff
@@ -9,10 +9,15 @@ import kotlin.reflect.KProperty
...
 public class Response<T>(
-	public val content: T,
+	public val content: T?,
 	public val status: Int,
 	public val headers: Map<String, List<String>>
 ) {
+	/**
+	 * Expect the [content] to be not-null.
+	 */
+	public val requireContent: T get() = content!!
```
Maxr1998, line 19: Don't know, should maybe be a function `requireContent()` or be called `requiredContent` or something.

nielsvanvelzen: Doing it as a property makes it easier to use in my opinion. Easier to write `xx.requireContent.yy`. Although I can understand why you want it as function. I don't have a preference for one or the other so you call it.

nielsvanvelzen: Maybe we don't need it at all? Because you can still use `by xx`.

```kotlin
val info = systemApi.getPublicInfo()

// a
val result by info

// b
val result = info.content!!

// c
val result = info.requireContent

// d
val result = info.requireContent()
```

Maxr1998: Both work fine, but with the current name it sounds more like a function. Properties do look neater in this case though, you're right - but then a different name should be used.

Maxr1998: > Maybe we don't need it at all? Because you can still use `by xx`.
> 
> ```kotlin
> val info = systemApi.getPublicInfo()
> 
> // a
> val result by info
> 
> // b
> val result = info.content!!
> 
> // c
> val result = info.requireContent
> 
> // d
> val result = info.requireContent()
> ```

Didn't see the updated comment yet, you're right - it isn't really necessary and can also be removed.

nielsvanvelzen: Removed the requireContent prop and added a `requireNotNull` check in the `getValue` function to better message developers of a bad request,

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
```diff
@@ -207,22 +214,39 @@ public open class KtorClient(
...
+			if (!response.status.isSuccess()) throw InvalidStatusError(response.status.value)
+			// Read response body
+			val body = response.receive<T>()
+			// Return custom response instance
+			return Response(body, response.status.value, response.headers.toMap())
+		} catch (err: HttpRequestTimeoutException) {
+			logger.debug("Connection timed out", err)
+			throw TimeoutError(err.message)
+		} catch (err: NoTransformationFoundException) {
+			logger.error("Requested model does not exist!?", err)
```
Maxr1998, line 244: `!?` is a bit aggressive, but sure :sweat_smile:

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/error/ApiClientError.kt
```diff
@@ -0,0 +1,8 @@
+package org.jellyfin.sdk.api.client.error
+
+/**
+ * Base exception for the ApiClient.
+ */
+public abstract class ApiClientError(
+	message: String?
+) : Error(message)
```
Maxr1998, line 8: Could be single-line. If not, should have a trailing comma. Also why `Error` and not `Exception`?
Same for the other error classes ofc.

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
```diff
@@ -207,22 +214,39 @@ public open class KtorClient(
...
+			throw InvalidContentError(err.message)
+		} catch (err: SerializationException) {
+			logger.error("Deserialization failed", err)
+			throw InvalidContentError(err.message)
+		}
```
Maxr1998, line 249: I really like that now :eyes:

Maxr1998 (approved):

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
nielsvanvelzen: If we used the Android logger this would be a `logger.wtf` because this should not be possible to happen.

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/error/ApiClientError.kt
nielsvanvelzen: Should be exception, I'll change it.

nielsvanvelzen: also, multi-line for readability

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
Maxr1998: Unrelated, but I still love that method name. *What a terrible failure* - yeah sure.

Maxr1998 (approved):

PR #255, opened 2021-05-14
Add pair extensions to easily convert SDK pairs to and from Kotlin pairs

nielsvanvelzen:
`Pair` -> `Model` & `Model` -> `Pair` for:

- NameIdPair
- NameGuidPair
- NameValuePair
- XmlAttribute


Maxr1998 (approved):
Maybe instead of `key` and `value` for the KDoc you could also write `first` and `second` so that it matches the attributes of `Pair`.

Maxr1998 (approved):

PR #256, opened 2021-05-14
Rename MissingPathVariableError to MissingPathVariableException and extend ApiClientException

nielsvanvelzen:
Forgot this in #246


jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/exception/MissingPathVariableException.kt
```diff
@@ -0,0 +1,9 @@
+package org.jellyfin.sdk.api.client.exception
+
+/**
+ * One or multiple variables in the path template are missing.
```
Maxr1998, line 4:
```suggestion
 * A variable in the path template is missing.
```

Maxr1998 (approved):

Maxr1998 (approved):

PR #257, opened 2021-05-14
Update kotlinx.coroutines and fix warnings in WebSocketApi

nielsvanvelzen:
I'm probably rewriting the WebSocketApi at some point, I'm not satisfied with the current implementation. The current changes are minimal just to fix the warnings after the update.


Maxr1998 (approved):
LGTM

PR #259, opened 2021-05-19
Catch all types of timeouts and rethrow as ApiClientException

Maxr1998:
Fixes jellyfin/jellyfin-android#368.


jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
```diff
@@ -239,8 +240,14 @@ public open class KtorClient(
 			// Return custom response instance
 			return Response(body, response.status.value, response.headers.toMap())
 		} catch (err: HttpRequestTimeoutException) {
+			logger.debug("Http request timed out", err)
```
nielsvanvelzen, line 243:
```suggestion
			logger.debug("HTTP request timed out", err)
```

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

PR #263, opened 2021-05-28
Add bitrate measurement functions

nielsvanvelzen:
Adds 2 new functions to the `MediaInfoApi` class:

- measureBitrate
- detectBitrate

Both return a `BitrateMeasurement` containing requested bytes, response duration and calculated bitrate.
The detect function should normally take between 0.5-3 seconds but can take longer depending on network conditions.

Additionally adds a new bitrate command to the CLI because why not.

### DIFF
--- a/jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
+++ b/jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
@@ -1,0 +1,45 @@
+package org.jellyfin.sdk.api.client.extensions
+
+import org.jellyfin.sdk.api.operations.MediaInfoApi
+import org.jellyfin.sdk.model.BitrateMeasurement
+
+/**
+ * Create a [BitrateMeasurement] for a response of [bytes]. The returned value contains the input, duration and the
+ * bitrate.
+ */
+public suspend fun MediaInfoApi.measureBitrate(bytes: Int): BitrateMeasurement {
+	val now = System.currentTimeMillis()
+
+	// Request the bits
+	getBitrateTestBytes(size = bytes).apply {
+		// Download full content
+		content.awaitContent()
+	}
+
+	// Calculate duration
+	val duration = System.currentTimeMillis() - now
+
+	// Return results
+	return BitrateMeasurement(
+		bytes = bytes,
+		duration = duration,
+	)
+}
+
+/**
+ * Detect the bitrate of the current connection by calling [measureBitrate] until a measurement of at least 1 second or
+ * with the maximum byte size is created.
+ */
+@Suppress("MagicNumber")
+public suspend fun MediaInfoApi.detectBitrate(): BitrateMeasurement {
+	val maxBytes = 10_000_000 // Max is 10 MB
+	var testBytes = 4_000_000 // Start with 4 MB
+
+	var measurement: BitrateMeasurement
+	do {
+		measurement = measureBitrate(testBytes)
+		testBytes += 2_000_000 // Increase by 2 MB
+	} while (measurement.duration <= 1_000 && measurement.bytes < maxBytes)
+
+	return measurement
+}
--- a/jellyfin-model/src/main/kotlin/org/jellyfin/sdk/model/BitrateMeasurement.kt
+++ b/jellyfin-model/src/main/kotlin/org/jellyfin/sdk/model/BitrateMeasurement.kt
@@ -1,0 +1,20 @@
+package org.jellyfin.sdk.model
+
+public data class BitrateMeasurement(
+	/**
+	 * Amount of bytes used to measure the bitrate.
+	 */
+	val bytes: Int,
+
+	/**
+	 * Measurement duration in milliseconds.
+	 */
+	val duration: Long,
+) {
+	/**
+	 * Measured network speed in bits per second.
+	 * Calculated by dividing [bytes] by the [duration] in seconds multiplied by 8 (byte -> bits).
+	 */
+	@Suppress("MagicNumber")
+	val bitrate: Long = bytes * 8000L / duration
+}
--- a/jellyfin-api/api/jellyfin-api.api
+++ b/jellyfin-api/api/jellyfin-api.api
@@ -102,6 +102,11 @@ public final class org/jellyfin/sdk/api/client/exception/TimeoutException : org/
 	public fun <init> (Ljava/lang/String;)V
 }
 
+public final class org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensionsKt {
+	public static final fun detectBitrate (Lorg/jellyfin/sdk/api/operations/MediaInfoApi;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public static final fun measureBitrate (Lorg/jellyfin/sdk/api/operations/MediaInfoApi;ILkotlin/coroutines/Continuation;)Ljava/lang/Object;
+}
+
 public final class org/jellyfin/sdk/api/client/extensions/UserApiExtensionsKt {
 	public static final fun authenticateUserByName (Lorg/jellyfin/sdk/api/operations/UserApi;Ljava/lang/String;Ljava/lang/String;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 }
--- a/jellyfin-model/api/jellyfin-model.api
+++ b/jellyfin-model/api/jellyfin-model.api
@@ -1,3 +1,17 @@
+public final class org/jellyfin/sdk/model/BitrateMeasurement {
+	public fun <init> (IJ)V
+	public final fun component1 ()I
+	public final fun component2 ()J
+	public final fun copy (IJ)Lorg/jellyfin/sdk/model/BitrateMeasurement;
+	public static synthetic fun copy$default (Lorg/jellyfin/sdk/model/BitrateMeasurement;IJILjava/lang/Object;)Lorg/jellyfin/sdk/model/BitrateMeasurement;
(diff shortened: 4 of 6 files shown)


jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
```diff
@@ -0,0 +1,61 @@
...
+ * @param minBitrate The resulting [BitrateMeasurement.bitrate] field is at least [minBitrate]. Defaults to 1Mb.
+ *
+ * @throws AssertionError If the server response is invalid.
+ */
+@Suppress("MagicNumber")
+public suspend fun MediaInfoApi.detectBitrate(
+	minBitrate: Long = 1_000_000, // 1Mb
+): BitrateMeasurement {
+	val maxBytes = 10_000_000 // Max is 10mb
+	var test = 4_000_000 // start with a response of 4mb
```
Maxr1998, line 49:
```suggestion
	val maxBytes = 10_000_000 // Max is 10 MB
	var test = 4_000_000 // start with a response of 4 MB
```

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
```diff
@@ -0,0 +1,61 @@
...
+
+	var measurement: BitrateMeasurement
+	do {
+		measurement = measureBitrate(test)
+		test += 2_000_000 // Increase with 2mb
```
Maxr1998, line 54:
```suggestion
		test += 2_000_000 // Increase with 2 MB
```

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
Maxr1998, line 49: Would be a better name imo.

```suggestion
	var testBytes = 4_000_000 // start with a response of 4mb
```

jellyfin-model/src/main/kotlin/org/jellyfin/sdk/model/BitrateMeasurement.kt
```diff
@@ -0,0 +1,19 @@
...
+	val bytes: Int,
+
+	/**
+	 * How long the measurement took in milliseconds.
+	 */
+	val duration: Long,
+
+	/**
+	 * Amount of bits in a second.
+	 * Calculated by dividing [bytes] by the [duration] multiplied by 1 second multiplied by 8 bits (byte->bit).
```
Maxr1998, line 16:
```suggestion
	 * Calculated by dividing [bytes] by the [duration] in seconds multiplied by 8 (byte → bits).
```

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
```diff
@@ -0,0 +1,61 @@
...
+		duration = duration,
+		bitrate = bytes / duration * 8000 // bytes -> bits
+	)
+}
+
+/**
+ * Detect the bitrate of the current connection by calling [measureBitrate] until a measurement of at least 1 second or
+ * with the maximum byte size is created.
+ *
+ * @param minBitrate The resulting [BitrateMeasurement.bitrate] field is at least [minBitrate]. Defaults to 1Mb.
```
Maxr1998, line 40: I don't understand what this is for, what would be a use-case for it?

Maxr1998 (changes requested):

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
nielsvanvelzen: - Prevent too low values in case of weird network conditions
- Set a minimal required bound for the result
- Allow the application to change the minimal value

The default value (1 Mb) is the bitrate similar to 480p h.264 video.

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @throws AssertionError If the server response is invalid.
+ */
+@Suppress("MagicNumber")
+public suspend fun MediaInfoApi.detectBitrate(
+	minBitrate: Long = 1_000_000, // 1 Mb
+): BitrateMeasurement {
+	val maxBytes = 10_000_000 // Max is 1 0MB
```
Maxr1998, line 48:
```suggestion
	val maxBytes = 10_000_000 // Max is 10 MB
```

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
```diff
@@ -0,0 +1,61 @@
...
+	var testBytes = 4_000_000 // start with a response of 4 MB
```
Maxr1998, line 49:
```suggestion
	var testBytes = 4_000_000 // Start with 4 MB
```

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
```diff
@@ -0,0 +1,61 @@
...
+
+	var measurement: BitrateMeasurement
+	do {
+		measurement = measureBitrate(testBytes)
+		testBytes += 2_000_000 // Increase with 2 MB
```
Maxr1998, line 54:
```suggestion
		testBytes += 2_000_000 // Increase by 2 MB
```

jellyfin-model/src/main/kotlin/org/jellyfin/sdk/model/BitrateMeasurement.kt
```diff
@@ -0,0 +1,19 @@
+package org.jellyfin.sdk.model
+
+public data class BitrateMeasurement(
+	/**
+	 * The requested bytes
+	 */
+	val bytes: Int,
+
+	/**
+	 * How long the measurement took in milliseconds.
```
Maxr1998, line 10:
```suggestion
	 * Measurement duration in milliseconds.
```

jellyfin-model/src/main/kotlin/org/jellyfin/sdk/model/BitrateMeasurement.kt
```diff
@@ -0,0 +1,19 @@
+package org.jellyfin.sdk.model
+
+public data class BitrateMeasurement(
+	/**
+	 * The requested bytes
+	 */
+	val bytes: Int,
```
Maxr1998, line 7:
```suggestion
	/**
	 * The amount of bytes used to measure the bitrate.
	 */
	val size: Int,
```

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
Maxr1998: Why should the SDK handle that, however? It even violates the bitrate = bytes/duration invariant for the measurement object..
I think it'd make more sense to move that decision to the client, and return the bitrate exactly like it was measured, however good or bad it was (and without caring about the quality of the measurement).
You could just add a comment to the KDoc suggesting using a `max(result, minimum)`.

Maxr1998 (changes requested):

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/extensions/MediaInfoApiExtensions.kt
```diff
@@ -0,0 +1,61 @@
+package org.jellyfin.sdk.api.client.extensions
+
+import org.jellyfin.sdk.api.operations.MediaInfoApi
+import org.jellyfin.sdk.model.BitrateMeasurement
+
+/**
+ * Create a [BitrateMeasurement] for a response of [bytes]. The returned value contains the input, duration and the
+ * bitrate.
+ *
+ * @throws AssertionError If the server response is invalid.
```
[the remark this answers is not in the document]
nielsvanvelzen: It's the exception `assert()` uses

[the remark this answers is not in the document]
Maxr1998: I know, but I'd personally prefer a custom exception here that extends `ApiClientException`.

jellyfin-model/src/main/kotlin/org/jellyfin/sdk/model/BitrateMeasurement.kt
```diff
@@ -0,0 +1,19 @@
...
+
+	/**
+	 * Measurement duration in milliseconds.
+	 */
+	val duration: Long,
+
+	/**
+	 * Amount of bits in a second.
```
Maxr1998, line 15:
```suggestion
	 * Amount of bits per second.
```

Maxr1998 (approved):

Maxr1998 (approved):

PR #264, opened 2021-05-28
Publish openapi-unstable branch as openapi-unstable-SNAPSHOT

nielsvanvelzen:
- Add jellyfin.branch property to create versions from a branch name
- Publish openapi-unstable branch as openapi-unstable-SNAPSHOT
- getProperty log set to debug
- Rename `latest-SNAPSHOT` to `master-SNAPSHOT`
  - latest is now a fallback which is not used in CI, update your projects using the snapshots to use `master-SNAPSHOT`!

This PR is untested since I can't test this on my fork.

### DIFF
--- a/buildSrc/src/main/kotlin/Properties.kt
+++ b/buildSrc/src/main/kotlin/Properties.kt
@@ -7,6 +7,17 @@ fun Project.getProperty(name: String): String? {
 	// sample.var --> SAMPLE_VAR
 	val environmentName = name.toUpperCase().replace(".", "_")
 	val value = findProperty(name)?.toString() ?: System.getenv(environmentName) ?: null
-	logger.info("getProperty($name): $environmentName - found=${!value.isNullOrBlank()}")
+	logger.debug("getProperty($name): $environmentName - found=${!value.isNullOrBlank()}")
 	return value
+}
+
+/**
+ * Helper function to create the SDK version using the `jellyfin.version` and `jellyfin.branch` properties.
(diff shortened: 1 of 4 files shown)


h1dden-da3m0n (approved):

buildSrc/src/main/kotlin/Properties.kt
```diff
@@ -7,6 +7,21 @@ fun Project.getProperty(name: String): String? {
...
+ */
+fun Project.createVersion(): String {
+	val jellyfinVersion = getProperty("jellyfin.version")?.removePrefix("v")
+	val jellyfinBranch = getProperty("jellyfin.branch")?.let { "$it-SNAPSHOT" }
+
+	return when {
+		jellyfinVersion != null -> jellyfinVersion
+		jellyfinBranch != null -> jellyfinBranch
+		else -> "latest-SNAPSHOT"
+	}
```
Maxr1998, line 26:
```suggestion
	return jellyfinVersion ?: jellyfinBranch ?: "latest-SNAPSHOT"
```

Maxr1998 (approved):

buildSrc/src/main/kotlin/Properties.kt
nielsvanvelzen: oh yeah like this makes sense actually, My first approach used the elvis operators with the `getProperty` calls

Maxr1998 (approved):

PR #268, opened 2021-06-05
Improve usage of Jellyfin class in Java code

nielsvanvelzen:
2 changes:

- Add additional constructor so you can pass the `JellyfinOptions.Builder` to the constructor (no need to call `.build()`)
- Add `@JvmOverloads` to the `createApi` function so you can call `jellyfin.createApi("baseurl")` instead of `jellyfin.createApi("baseurl", null, null, null, null)`


nielsvanvelzen:
Sample usage: https://github.com/nielsvanvelzen/jellyfin-sdk-kotlin/blob/3a47eb29e32a3d3afe623923561b5b5c806b2283/samples/java-cli/src/main/java/org/jellyfin/sample/cli/Main.java

Maxr1998 (approved):

PR #269, opened 2021-06-07
Add toString function to ServerVersion

nielsvanvelzen:
Required for jellyfin-androidtv

### DIFF
--- a/jellyfin-model/src/main/kotlin/org/jellyfin/sdk/model/ServerVersion.kt
+++ b/jellyfin-model/src/main/kotlin/org/jellyfin/sdk/model/ServerVersion.kt
@@ -16,6 +16,19 @@ public data class ServerVersion(
 ) {
 	public operator fun compareTo(other: ServerVersion): Int = comparator.compare(this, other)
 
+	/**
+	 * Convert version to string. Format is "[major].[minor].[patch].[build]".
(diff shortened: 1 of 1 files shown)


jellyfin-model/src/main/kotlin/org/jellyfin/sdk/model/ServerVersion.kt
```diff
@@ -16,6 +16,19 @@ public data class ServerVersion(
...
+	 * Convert version to string. Format is "[major].[minor].[patch].[build]".
+	 * [build] is omitted if null.
+	 * Sample output:
+	 * - 1.0.0
+	 * - 10.6.4
+	 * - 10.7.0.0
+	 */
+	override fun toString(): String = buildString {
+		append(major, '.', minor, '.', patch)
+		if (build != null) append('.', build)
```
Maxr1998, line 29: Why no String templates + an if?

nielsvanvelzen: Slightly easier to read this way imo

Maxr1998: Fair enough

Maxr1998 (approved):

PR #270, opened 2021-06-09
Add DefaultValueHook for setting default operation parameter values

nielsvanvelzen:
Will be used in my next pull requests that automagically adds the userId as parameter.


jellyfin-api/src/main/kotlin-generated/org/jellyfin/sdk/api/operations/MediaInfoApi.kt
```diff
@@ -86,7 +86,7 @@ public class MediaInfoApi(
 
 	/**
 	 * For backwards compatibility parameters can be sent via Query or Body, with Query having higher
-	 * precedence.
+	 * precedence.
```
nielsvanvelzen, line 89: I have absolutely no idea why this keeps happening, I guess it's a Windows<>Linux thing?

Maxr1998: Tabs vs spaces maybe?

nielsvanvelzen: nop, no whitespace changes

dkanada: Perhaps the CRLF character has changed?

nielsvanvelzen: But why wouldn't that happen in other files? All of them are regenerated

Maxr1998: > Perhaps the CRLF character has changed?

That in fact seems to be the case - CRLF (Windows) gets changed to LF (UNIX).

Git diff is
`09 20 2a 20 70 72 65 63 65 64 65 6e 63 65 2e 0d 0a`
vs
`09 20 2a 20 70 72 65 63 65 64 65 6e 63 65 2e 0a`

openapi-generator/README.md
```diff
@@ -61,6 +61,11 @@ Hooks are classes that will modify the output of the generator. They should be r
     used to fix spelling, moving operations to different services or adding operations to additional
     services.
 
+  - **DefaultValueHook**  
+    A hook that allows for changing the default value of parameters in operations. The hook returns
```
Maxr1998, line 65:
```suggestion
    A hook that allows changing the default value of parameters in operations. The hook returns
```

Maxr1998, line 65:
```suggestion
    A hook that allows changing the default value of parameters in operations. The hook returns
```

openapi-generator/README.md
```diff
@@ -61,6 +61,11 @@ Hooks are classes that will modify the output of the generator. They should be r
...
+    a class that contains the code builder. Generator will use the default value from the JSON
```
Maxr1998, line 66:
```suggestion
    a class that contains the code builder. The generator will use the default value from the JSON
```

openapi-generator/README.md
```diff
@@ -61,6 +61,11 @@ Hooks are classes that will modify the output of the generator. They should be r
...
+    schema when all hooks return null.
```
Maxr1998, line 67:
```suggestion
    schema if all hooks return null.
```

openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
```diff
@@ -13,17 +15,16 @@ import org.jellyfin.openapi.constants.MimeType
 import org.jellyfin.openapi.constants.Security
 import org.jellyfin.openapi.constants.Strings
 import org.jellyfin.openapi.hooks.ApiTypePath
+import org.jellyfin.openapi.hooks.DefaultValueHook
 import org.jellyfin.openapi.hooks.ServiceNameHook
-import org.jellyfin.openapi.model.ApiService
-import org.jellyfin.openapi.model.ApiServiceOperation
-import org.jellyfin.openapi.model.ApiServiceOperationParameter
-import org.jellyfin.openapi.model.HttpMethod
+import org.jellyfin.openapi.model.*
```
Maxr1998, line 20: Doesn't the style enforce no wildcards?

Maxr1998 (changes requested):

openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
nielsvanvelzen: yes but intellij/android studio doesn't

Maxr1998: Well, you can turn them off in the IDE, but unfortunately it's not that reliable..

nielsvanvelzen: If the EditorConfig file can't change those I won't use it, because not all contributors will.

openapi-generator/README.md
```diff
@@ -61,6 +61,11 @@ Hooks are classes that will modify the output of the generator. They should be r
     used to fix spelling, moving operations to different services or adding operations to additional
     services.
 
+  - **DefaultValueHook**  
+    A hook that allows changing the default value of parameters in operations. The hook returns
+    a class that contains the code builder. The generator will use the default value from the JSON
```
Maxr1998, line 66: Maybe a bit more specific.

```suggestion
    an instance of `CustomDefaultValue` that contains the code builder. The generator will use the default value from the JSON
```

Maxr1998 (dismissed):

openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
Maxr1998: I know, I was mostly talking about this case, since it's an unnecessary change and we could at least enforce it in reviews and for ourselves, even if it doesn't work for all contributors by default.

nielsvanvelzen: Fair enough, I think in this case it changed to wildcard because it now uses 5>= imports from the model package

Maxr1998 (approved):

PR #272, opened 2021-06-12
Add userId default value hook

nielsvanvelzen:
This new hook will use the value of `ApiClient.userId` by default for all API operations that require a userId (body models and nullables not supported).


jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/exception/UserIdIsNullException.kt
```diff
@@ -0,0 +1,5 @@
+package org.jellyfin.sdk.api.client.exception
+
+public class UserIdIsNullException : ApiClientException(
```
Maxr1998, line 3:
```suggestion
public class MissingUserIdException : ApiClientException(
```

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/ApiClient.kt
```diff
@@ -22,6 +23,12 @@ public interface ApiClient {
 	 */
 	public var accessToken: String?
 
+	/**
+	 * User id automatically used in API operations that require it.
+	 * Should be the same id as the user for [accessToken].
```
Maxr1998, line 28:
```suggestion
	 * User ID that will automatically be used in user-specific API operations.
	 * Should correspond to the same user as [accessToken].
```

Maxr1998 (changes requested):

Maxr1998 (approved):

PR #274, opened 2021-06-13
Fix bad line endings used in openapi generator

nielsvanvelzen:
This was the easiest way to fix it, although not the prettiest.


openapi-generator/src/main/kotlin/org/jellyfin/openapi/Main.kt
```diff
@@ -12,7 +12,7 @@ fun main(vararg args: String) {
 	val modelsOutputDir by arguments
 
 	// Read OpenAPI json
-	val openApiJson = File(openApiFile).readText()
+	val openApiJson = File(openApiFile).readText(Charsets.UTF_8)
```
Maxr1998, line 15: That's the default, so theoretically not needed.

Maxr1998 (dismissed):

openapi-generator/src/main/kotlin/org/jellyfin/openapi/Main.kt
nielsvanvelzen: Oh that's a leftover from testing let me remove it

Maxr1998 (approved):

PR #278, opened 2021-06-16
Add simplified Java version of the cli sample

nielsvanvelzen:
It doesn't contain commands etc. but it does show how to use the API in Java. Can be extended in the future.


nielsvanvelzen:
no squashing so the Java8+ sources are in the git history

Maxr1998 (approved):

PR #287, opened 2021-06-30
Fix formatting of LocalDateTimeSerializer encoding and add tests

### LINKED ISSUE #285: LocalDateTime can't serialized to json
I try to use *ItemUpdateApi#updateItem(itemId: UUID, `data`: BaseItemDto)*

but an error occurred

```json
{
   "type":"https://tools.ietf.org/html/rfc7231#section-6.5.1",
   "title":"One or more validation errors occurred.",
   "status":400,
   "traceId":"00-33dd05cfdb47ed4088bf0ae90b8b86da-8b9f9586acc83544-00",
   "errors":{
      "request":[
         "The request field is required."
      ],
      "$.DateCreated":[
         "The JSON value could not be converted to System.DateTime. Path: $ | LineNumber: 0 | BytePositionInLine: 52. Path: $.DateCreated | LineNumber: 0 | BytePositionInLine: 268."
      ]
   }
}
```

nielsvanvelzen:
Should fix #285 although not directly tested with the api operation mentioned in the issue.

### DIFF
--- a/jellyfin-model/api/jellyfin-model.api
+++ b/jellyfin-model/api/jellyfin-model.api
@@ -11055,6 +11055,8 @@ public final class org/jellyfin/sdk/model/extensions/PairExtensionsKt {
 
 public final class org/jellyfin/sdk/model/serializer/LocalDateTimeSerializer : kotlinx/serialization/KSerializer {
 	public fun <init> ()V
+	public fun <init> (Ljava/time/ZoneId;)V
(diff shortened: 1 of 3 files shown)


Maxr1998:
Should maybe ensure that ZoneId can be desugared on Android before merging.

nielsvanvelzen:
java.time is available with desugaring, it's exactly why it was added

https://developer.android.com/studio/write/java8-support-table

Maxr1998:
I suppose not all parts of the package are supported since they explicitly list classes and methods, but the aforementioned one is, so it should be fine.

Maxr1998 (approved):

PR #289, opened 2021-07-08
Miscellaneous build script changes

nielsvanvelzen:
Makes the SDK build scripts more consistent with https://github.com/jellyfin/jellyfin-androidtv/pull/1001 (which was made to be more consistent with the SDK??)


android-lint.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<lint>
+</lint>
```
Maxr1998, line 3: Why add it when it's empty?

Maxr1998 (approved):

android-lint.xml
nielsvanvelzen: If we do configure issues in the future the file is already in the correct place

PR #292, opened 2021-07-09
Add ParameterValidation framework with initial IntRangeValidation

nielsvanvelzen:
Allows for client-side checking of values based on the metadata available in the OpenAPI specification. Right now (with 10.7.x) none of the parameters add any validation metadata.

Sample of generated code with the changes from https://github.com/jellyfin/jellyfin/pull/6276

```kotlin
	/**
	 * Tests the network with a request with the size of the bitrate.
	 *
	 * @param size The bitrate. Defaults to 102400.
	 */
	public suspend fun getBitrateTestBytes(size: Int? = 102400): Response<ByteReadChannel> {
		val pathParameters = emptyMap<String, Any?>()
		val queryParameters = mutableMapOf<String, Any?>()
		require(size in 0..100000000) { "Parameter \"size\" must be in range 0..100000000 (inclusive)." }
		queryParameters["size"] = size
		val data = null
		val response = api.`get`<ByteReadChannel>("/Playback/BitrateTest", pathParameters,
				queryParameters, data)
		return response
	}
```

### DIFF
--- a/openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/api/OperationBuilder.kt
+++ b/openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/api/OperationBuilder.kt
@@ -8,9 +8,7 @@ import org.jellyfin.openapi.builder.extra.DescriptionBuilder
 import org.jellyfin.openapi.constants.Classes
 import org.jellyfin.openapi.constants.Packages
 import org.jellyfin.openapi.constants.Strings
-import org.jellyfin.openapi.model.ApiServiceOperation
-import org.jellyfin.openapi.model.ApiServiceOperationParameter
-import org.jellyfin.openapi.model.CustomDefaultValue
+import org.jellyfin.openapi.model.*
 
 open class OperationBuilder(
 	private val descriptionBuilder: DescriptionBuilder,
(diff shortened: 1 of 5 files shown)


nielsvanvelzen:
Ideally I'd add an annotation too, like `@IntRange` but unfortunately there is no such thing in Kotlin stdlib. Can't use the androidx.annotation library either as that would require the google repo for non-android projects 😢

crobibero:
The minimum size is 1 not 0 :)

nielsvanvelzen:
I manually changed the JSON file to test it so it should work fine with the generated file 👍

Maxr1998:
> Sample of generated code with the changes from https://github.com/jellyfin/jellyfin/pull/6276
> 
> ```kotlin
>   …
> 		require(size in 0..100000000) { "Parameter \"size\" must be in range 0..100000000 (inclusive)." }
> ```

Shouldn't it be `require(size in 1..100000000) { "Parameter \"size\" must be in range 1..100000000 (inclusive)." }` in this case?

nielsvanvelzen:
> Shouldn't it be
> 
> ```kotlin
> require(size in 1..100000000) { "Parameter \"size\" must be in range 1..100000000 (inclusive)." }
> ```
> 
> in this case?


See my other comment:

> I manually changed the JSON file to test it so it should work fine with the generated file 👍

Maxr1998 (approved):

PR #293, opened 2021-07-09
Correctly set member names in OperationBuilder

nielsvanvelzen:
- Use %M instead of %N so the imports are properly created

No real significant changes except for the fact that it adds imports for stuff that is imported by default.

### DIFF
--- a/jellyfin-api/src/main/kotlin-generated/org/jellyfin/sdk/api/operations/ActivityLogApi.kt
+++ b/jellyfin-api/src/main/kotlin-generated/org/jellyfin/sdk/api/operations/ActivityLogApi.kt
@@ -10,6 +10,8 @@ import kotlin.Any
 import kotlin.Boolean
 import kotlin.Int
 import kotlin.String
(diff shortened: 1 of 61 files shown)


Maxr1998:
Why are those necessary if they are imported by default? Not really getting the point of this PR?

nielsvanvelzen:
It fixes an incorrect usage of KotlinPoet templating (using %N instead of %M) which causes the imports to be added. KotlinPoet adds all usages of stdlib as imports to make sure no naming conflicts arise (adding a function called `emptyList` in the `org.jellyfin.androidtv.api.operations` package would use that one instead of the stdlib one if there is no import, see https://github.com/square/kotlinpoet/issues/763).

The change is mostly to be consistent with the recommended usage of KotlinPoet and doesn't change any behavior.

Maxr1998:
I see, so it's for future proofing and robustness against changes. Makes sense.

Maxr1998 (approved):

PR #295, opened 2021-07-11
Ensure device name has no special characters

Nazar78:
The below checks for "special characters" on client level wasn't enough to cause the client not being able to connect to the server when it has special characters in the device name. Proposed to have checks at SDK level instead. Additionally to fallback to use the manufacturer and model string instead if the device name has special characters:

https://github.com/jellyfin/jellyfin-android/blob/0d2457cf660f6ba2cd9c4013f7afdd044b357f0d/app/src/main/java/org/jellyfin/mobile/bridge/NativeInterface.kt#L63

### DIFF
--- a/jellyfin-platform-android/src/main/kotlin/org/jellyfin/sdk/interaction/AndroidDevice.kt
+++ b/jellyfin-platform-android/src/main/kotlin/org/jellyfin/sdk/interaction/AndroidDevice.kt
@@ -10,9 +10,21 @@ import org.jellyfin.sdk.model.DeviceInfo
 public fun androidDevice(context: Context): DeviceInfo {
 	val id = Settings.Secure.getString(context.contentResolver, Settings.Secure.ANDROID_ID)
 
+	// Ensure device name has no special characters or fallback to manufacturer and model else the connection will fail
+	val testDeviceName = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
+		val deviceName = Settings.Global.getString(context.contentResolver, Settings.Global.DEVICE_NAME)
+		if (deviceName.matches("([^\\x20-\\x7E])".toRegex())) {
+			deviceName
+		} else {
+			null
+		}
+	} else {
+		null
+	}
+
 	// Use name from device settings
-	val name = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
-		Settings.Global.getString(context.contentResolver, Settings.Global.DEVICE_NAME)
+	val name = if (testDeviceName != null) {
+		testDeviceName
 	} else {
 		// Concatenate the name based on manufacturer and model
 		val manufacturer = Build.MANUFACTURER


nielsvanvelzen:
Can you explain what you're trying to achieve with this change? The mobile android app needs to sanitize the device name when passing it to the webview but the SDK itself doesn't need that. The SDK is able to properly send special characters to the server. Other clients, like the Android TV app, make use of this. An example of the Android TV app that sends special characters in the device name via the SDK:

image

Nazar78:
Even with sanitization done from the mobile android beta app, it doesn't seems to work. I've also tried changing the codes on the android mobile app master to `put("deviceName", "test")` and with special characters it still won't connect. Only when I change it in the SDK then it will connect. This is also the same case when I tested with the android emulator.

I also have android TV connected but they have no issues unlike the android mobile app. The play store version also has no issues. I was thinking if we need to sanitize the device name, why not do it from the SDK so both android tv and android mobile app can make use of it?

Please refer to the screenshot below. This emulator has it's device name to have special characters `^-~`. If I use the default SDK version, it won't connect. But once I switched to locally edited SDK which is actually this PR, I can connect successfully.

image

nielsvanvelzen:
The name of my mobile emulator is `Emulator"=¶✓c` (has been like that for months now) and even when I change it to `Emulator"=¶✓c^-~` I can connect just fine to my own server. I tested this with 3 apps that use this SDK (jellyfin-android, jellyfin-androidtv and a private app).

Nazar78:
> Emulator"=¶✓c

Sorry for not being clear.  When I wrote `^-~` I meant regex for other characters not in range of `-` and `~`. Anyway I've tested your emulator name `Emulator"=¶✓c` it connects successfully but these characters aren't enough because when I append some other special characters it won't connect. Once I change to this PR, it connects. However at your convenience can you please help to replicate the following steps see if you can connect which I can't?

1. Using the following version which you probably already have:
- jellyfin-android master
- jellyfin-sdk-kotlin master
- jellyfin server 10.7.6
- AVD Pixel 2 API 29

2. Clear the Jellyfin Debug app storage or just uninstall it from the emulator.

3. Rename your emulator device name at the system settings to e.g. my name `Ñážàŕ`.

4. Run the debug and test if you can connect to your local server.

nielsvanvelzen:
Thanks, it appears that some special characters were in fact allowed. I can now reproduce and this is the exception I got:

```
    java.lang.IllegalArgumentException: Unexpected char 0xcb at 132 in Authorization value: MediaBrowser Client="Jellyfin Android", Version="0.0.0-dev.1", DeviceId="xxx", Device="xxx", Token="xxx"
        at okhttp3.Headers$Companion.checkValue(Headers.kt:434)
        at okhttp3.Headers$Companion.access$checkValue(Headers.kt:346)
        at okhttp3.Headers$Builder.add(Headers.kt:245)
        at okhttp3.Request$Builder.addHeader(Request.kt:210)
        at io.ktor.client.engine.okhttp.OkHttpEngineKt$convertToOkHttpRequest$1$1.invoke(OkHttpEngine.kt:189)
        at io.ktor.client.engine.okhttp.OkHttpEngineKt$convertToOkHttpRequest$1$1.invoke(OkHttpEngine.kt:186)
        at io.ktor.client.engine.UtilsKt$mergeHeaders$2.invoke(Utils.kt:37)
        at io.ktor.client.engine.UtilsKt$mergeHeaders$2.invoke(Utils.kt:32)
        at io.ktor.util.StringValuesImpl.forEach(StringValues.kt:146)
        at io.ktor.client.engine.UtilsKt.mergeHeaders(Utils.kt:32)
        at io.ktor.client.engine.okhttp.OkHttpEngineKt.convertToOkHttpRequest(OkHttpEngine.kt:186)
        at io.ktor.client.engine.okhttp.OkHttpEngineKt.access$convertToOkHttpRequest(OkHttpEngine.kt:1)
        at io.ktor.client.engine.okhttp.OkHttpEngine.execute(OkHttpEngine.kt:72)
        at io.ktor.client.engine.HttpClientEngine$executeWithinCallContext$2.invokeSuspend(HttpClientEngine.kt:85)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

The check in OkHTTP uses the following check: `c == '\t' || c in '\u0020'..'\u007e'`. I'm now convinced we should, indeed, add a fix to the SDK.

Your current implementation is not how I'd fix it though. It fixes the automatically retrieved android device name but a client can still set a name manually and trigger this exception. It would be better to sanitize the name in the http layer (in the KtorClient class) instead. This way special characters are always removed no matter how they are inputted.

Nazar78:
> It fixes the automatically retrieved android device name but >>>a client can still set a name manually<<< and trigger this exception.

Can the android mobile or the android tv client set the device name manually in the app itself? I didn't know the apps can do this. Sorry I just started playing around with codes. so I thought it's always automatically retrieved to whatever that's set in the device system settings. Unless of course it's a custom app.

> It would be better to sanitize the name in the http layer (in the KtorClient class) instead.

Agreed to this however...

IMHO I don't really like the current android mobile sanitization because it cuts off or removed the unicode or special characters then in the server dashboard it'll display an incomplete device name e.g. for my case `Ñážàŕ's Note20 Ultra..._\m/` sanitized to `s Note20 Ultra_m`. Not my device alone, I have several remote users connecting to my server with such cuttoff names and it looks weird. So I thought it would be better to replace the device name with model/manufacturer instead if there's special characters that would cause this exception.

Let me know if you agree with this, that's to replace the device name with model/manufacturer instead of removing those special characters? I can close this PR and open a new one for the KtorClient similar to this.

nielsvanvelzen:
> Can the android mobile or the android tv client set the device name manually in the app itself? 

Nope

>I thought it's always automatically retrieved to whatever that's set in the device system settings. Unless of course it's a custom app.

This is the answer, the SDK is meant for general use and while don't allow custom names in our official apps that doesn't mean third-party clients won't.

> in the server dashboard it'll display an incomplete device name 

Unfortunately this is not something we can do anything about without changes to the server. I've advocated for that before but it's a drastic change and nobody volunteered to properly fix it. Ideally we'd be able to send urlencoded values in the header to allow special characters.

> So I thought it would be better to replace the device name with model/manufacturer instead if there's special characters that would cause this exception.

While I understand your point we prefer to use an actual device name instead of model/manufacturer because it's more clear. Even when some characters are removed. The majority of our users have device names without special characters (you're the first with this issue while the mobile app has been in beta for months now). I still prefer to use the solution proposed above: add the escaping to KtorClient and remove that when the server can properly read the special characters via urlencoding or something else.

PR #298, opened 2021-07-29
Update Android Gradle Plugin to 7.0.0

nielsvanvelzen:


nielsvanvelzen:
Hmm, guess we'll need to drop Java 8 support for this one.

Maxr1998:
Not a problem imo. Next LTS release will be out soon, too.

README.md
```diff
@@ -37,7 +37,7 @@
 ---
 
 The Jellyfin Kotlin SDK is a library implementing the Jellyfin API to easily access servers. It is currently available
-for the JVM and Android 4.4 and up. Java 8 or higher is required. Android versions below Android 8 should use
+for the JVM and Android 4.4 and up. Java 11 or higher is required. Android versions below Android 8 should use
```
Maxr1998, line 40: That's a bit misleading. Only the host PC building the app needs JDK 11 to support the new Gradle plugin, the bytecode target can still be 52.0 / Java version 8. I'm not even sure the Android Runtime supports bytecode with a version higher than 52.0 - that's why there's desugaring.

Maxr1998 (changes requested):

README.md
nielsvanvelzen: So while using Java 8 for building works right now it might not be for long. We might want to target Java 11 bytecode in Kotlin (supported in AGP7 https://developer.android.com/studio/releases/gradle-plugin#java-11) and we don't support it anymore for Android so it's easier to also drop support for JVM now. Java 8 is near end of life anyway.

I can maybe change it to this though:
```suggestion
for the JVM and Android 4.4 and up. Java 11 or higher is required for building. Android versions below Android 8 should use
```

Maxr1998: Honestly, as long as we don't set the compileOptions to Java 11, I wouldn't change the README at all.
If we do that, it should only read that JDK 11 is necessary for building. When targeting old Android releases, the AGP desugars the bytecode anyway.

Maxr1998 (approved):

PR #303, opened 2021-08-14
Replace slf4j with kotlin-logging

nielsvanvelzen:
Kotlin-logging still uses slf4j under the hood so nothing changes for our apps, just add sl4fj-simple as dependency to see the logs from the SDK like always.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,6 @@
 [versions]
 kotlin = "1.5.10"
+kotlinlogging = "2.0.10"
 kotlinx-coroutines = "1.5.0"
 kotlinx-serialization = "1.2.1"
 ktor = "1.6.0"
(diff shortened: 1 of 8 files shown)


gradle/libs.versions.toml
```diff
@@ -1,5 +1,6 @@
 [versions]
 kotlin = "1.5.10"
+kotlinlogging = "2.0.10"
```
nielsvanvelzen, line 3: Can't use `kotlin-loggin` as name until Gradle 7.2 unfortunately.

jellyfin-api/src/main/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
```diff
@@ -104,12 +104,14 @@ public open class KtorClient(
 		queryParameters: Map<String, Any?> = emptyMap(),
 		requestBody: Any? = null,
 	): Response<T> {
-		val logger = LoggerFactory.getLogger(this::class.java)
 		val url = createUrl(pathTemplate, pathParameters, queryParameters)
 
 		// Log HTTP call with access token removed
-		val safeUrl = accessToken?.let { url.replace(it, "******") } ?: url
-		logger.info("$method $safeUrl")
+		val logger = KotlinLogging.logger {}
```
Maxr1998, line 110: What's the reason for having the logger within the method here?

nielsvanvelzen: To not expose the logger outside of the inline function

Maxr1998 (approved):

PR #308, opened 2021-08-20
Convert jellyfin-core and jellyfin-platform-android to Kotlin Multiplatform

### LINKED ISSUE #208: Use Kotlin Multiplatform and merge platform-android with core
Moving the structure of the library to Kotlin Mutliplatform allows us to add the Android specific code to the core library. This would make it easier to use the library as users don't need to include the Android platform separately. It also makes it easier for us to support different platforms like JavaScript, WASM, Native or iOS in the future.

I already tested this in #187 and it might be a relatively easy change.

nielsvanvelzen:
This PR migrates the jellyfin-core module to Kotlin Multiplatform (kmpp). This changes the file structure slightly to separate JVM specific code. This change allows us to be futureproof and potentially support multiple platforms (web, ios etc).

Other modules will be in separate pull requests.

Closes #208

----

Android module is now merged into jellyfin-core.


jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -111,27 +111,27 @@ public class RecommendedServerDiscovery(
 		)
 		val api = SystemApi(client)
 
-		val info: Result<Response<PublicSystemInfo>>
-		val responseTime = measureTimeMillis {
```
Maxr1998, line 115: [That method](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.system/measure-time-millis.html) is not common??

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/JellyfinOptions.kt
```diff
@@ -1,32 +1,23 @@
...
-		}
-	}
+	public companion object
+}
+
+public inline fun createJellyfinOptions(
+	init: JellyfinOptions.Builder.() -> Unit,
+): JellyfinOptions = JellyfinOptions.Builder().run {
+	init()
+	build()
```
Maxr1998, line 22:
```suggestion
): JellyfinOptions = JellyfinOptions.Builder().apply(init).build()
```

jellyfin-core/src/androidMain/kotlin/org/jellyfin/sdk/JellyfinOptions.kt
```diff
@@ -0,0 +1,29 @@
...
+		public var clientInfo: ClientInfo? = null
+
+		public actual fun build(): JellyfinOptions {
+			requireNotNull(context) { "An Android context is required when using the jellyfin-android platform." }
+
+			return JellyfinOptions(
+				context!!,
+				clientInfo,
+				deviceInfo = androidDevice(context!!)
+			)
```
Maxr1998, line 24: Could even use an expression body afterwards.

```suggestion
			return JellyfinOptions(
				requireNotNull(context) { "An Android context is required when using the jellyfin-android platform." },
				clientInfo,
				deviceInfo = androidDevice(context!!)
			)
```

jellyfin-core/src/jvmCommonMain/kotlin/org/jellyfin/sdk/discovery/LocalServerDiscovery.kt
```diff
@@ -0,0 +1,131 @@
...
+	 * Send our broadcast message to a given address
+	 */
+	private fun discoverAddress(socket: DatagramSocket, address: InetAddress): Unit = try {
+		val message = DISCOVERY_MESSAGE.toByteArray()
+		val packet = DatagramPacket(message, message.size, address, DISCOVERY_PORT)
+		socket.send(packet)
+
+		logger.debug { "Discovering via $address" }
+	} catch (err: IOException) {
+		logger.error(err) { "Unable to send discovery message to $address" }
```
Maxr1998, line 51: Don't all these `{ "..." }` logging calls have more overhead since they need to execute another function? I'd only use them if we really need the lazy evaluation honestly.

Maxr1998 (changes requested):

jellyfin-core/src/main/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
nielsvanvelzen: Only jvm and native

jellyfin-core/src/jvmCommonMain/kotlin/org/jellyfin/sdk/discovery/LocalServerDiscovery.kt
nielsvanvelzen: They are only evaluated if the logging level is enabled (which it is most of the time) but there are no functions to just pass a string in kotlin-logging common.

Maxr1998: Weird decision for their API. Not a fan of using lazy evaluation everywhere even if we have logging enabled generally, but it looks like I have to accept that.

Maxr1998 (approved):

PR #314, opened 2021-08-28
Move ApiClient HTTP methods to extension functions

### LINKED ISSUE #207: Make KtorClient private and depend on ApiClient in generated code
Although we do use the ApiClient interface in most places it doesn't include the request, get, post and delete functions. This is because we use inline functions with reified types, which are not supported when using interfaces. Preferably we'd fix this by passing the class as parameters instead of using generics but when I looked into this it didn't look like a possibility.

Separating the KtorClient (implementation) from the ApiClient (interface) it would allow the usage of custom HTTP clients and make it easier to change it if we decide to drop Ktor at some point.

nielsvanvelzen:
This PR moves the get, post and delete functions from the KtorClient to extension functions on ApiClient (although casted to KtorClient right now..). This is the initial change needed to fix #207 and most code is not dependent on the KtorClient implementation anymore. The ApiClient interface is now used as much as possible.

Part of #207

### DIFF
--- a/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/extensions/HttpMethodExtensions.kt
+++ b/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/extensions/HttpMethodExtensions.kt
@@ -1,0 +1,45 @@
+package org.jellyfin.sdk.api.client.extensions
+
+import io.ktor.http.*
+import org.jellyfin.sdk.api.client.ApiClient
+import org.jellyfin.sdk.api.client.KtorClient
+import org.jellyfin.sdk.api.client.Response
+
+public suspend inline fun <reified T> ApiClient.get(
+	pathTemplate: String,
+	pathParameters: Map<String, Any?> = emptyMap(),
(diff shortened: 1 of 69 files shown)


jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/extensions/HttpMethodExtensions.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.jellyfin.sdk.api.client.KtorClient
+import org.jellyfin.sdk.api.client.Response
+
+public suspend inline fun <reified T> ApiClient.get(
+	pathTemplate: String,
+	pathParameters: Map<String, Any?> = emptyMap(),
+	queryParameters: Map<String, Any?> = emptyMap(),
+	requestBody: Any? = null,
+): Response<T> = (this as KtorClient).request(
+	method = HttpMethod.Delete,
```
nielsvanvelzen, line 14: Damn you're stupid

Maxr1998 (approved):

PR #319, opened 2021-09-02
Add url and path building to new UrlBuilder

nielsvanvelzen:
- This simplifies the ApiClient implementation making it easier to create your own etc.
- Moved the ApiClient.createUrl fun to the interface and use UrlBuilder there, implementations can overwrite it if they desire to do so
- Also added few additional tests


jellyfin-api/src/jvmMain/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
```diff
@@ -105,7 +65,7 @@ public actual open class KtorClient actual constructor(
 		queryParameters: Map<String, Any?>,
 		requestBody: Any?,
 	): Response<T> {
-		val url = createUrl(pathTemplate, pathParameters, queryParameters)
+		val url = UrlBuilder.buildUrl(requireNotNull(baseUrl), pathTemplate, pathParameters, queryParameters)
```
Maxr1998, line 68: Shouldn't this still use `createUrl` but from the ApiClient interface?

nielsvanvelzen: Hmm actually it should, I removed createUrl at first but added it back since the api's url functions use it

Maxr1998 (approved):

PR #322, opened 2021-09-16
Separate content (de)serialization from ApiClient

### LINKED ISSUE #207: Make KtorClient private and depend on ApiClient in generated code

nielsvanvelzen:
- Add new class "RawResponse"
  - ApiClient is now expected to return an instance of RawResponse
  - Uses the new ApiSerializer to read response from RawResponse
- Added ApiClient.request to ApiClient interface
  - no longer inline because of the RawResponse class
- Removed (de)serialization responsibility from Ktor(Client)
- Made the HttpClient in KtorClient private
- Use new ApiSerialization in WebSocketApi

Tested with a private app that authenticates (post body), retrieves library items (response body) and downloads (response body bytestream reading) and it worked without any issues.

This PR closes #207 - It doesn't make KtorClient private but allows custom ApiClient instances now. In the future I'd like to move the KtorClient to a separate Gradle module so it can be omitted for advanced usage of the SDK.

### DIFF
--- a/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/HttpMethod.kt
+++ b/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/HttpMethod.kt
@@ -1,0 +1,7 @@
+package org.jellyfin.sdk.api.client
+
+public enum class HttpMethod {
+	GET,
+	POST,
+	DELETE,
+}
--- a/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/util/ApiSerializer.kt
+++ b/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/util/ApiSerializer.kt
@@ -1,0 +1,26 @@
+package org.jellyfin.sdk.api.client.util
+
+import io.ktor.utils.io.*
+import kotlinx.serialization.*
+import kotlinx.serialization.json.Json
+
+@OptIn(ExperimentalSerializationApi::class, InternalSerializationApi::class)
+public object ApiSerializer {
+	public val json: Json = Json {
+		isLenient = false
+		ignoreUnknownKeys = true
+		allowSpecialFloatingPointValues = true
+		useArrayPolymorphism = false
+	}
+
+	public fun encodeRequestBody(requestBody: Any? = null): String? {
+		if (requestBody == null) return null
+
+		return json.encodeToString(requestBody::class.serializer() as KSerializer<Any>, requestBody)
+	}
+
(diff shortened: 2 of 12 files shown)


jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/HttpMethod.kt
```diff
@@ -0,0 +1,7 @@
+package org.jellyfin.sdk.api.client
+
+public enum class HttpMethod {
+	GET,
+	POST,
+	DELETE
```
Maxr1998, line 6:
```suggestion
	DELETE,
```

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/util/ApiSerializer.kt
```diff
@@ -0,0 +1,29 @@
...
+		useArrayPolymorphism = false
+	}
+
+	public fun encodeRequestBody(requestBody: Any? = null): String? {
+		if (requestBody == null) return null
+
+		return json.encodeToString(
+			requestBody::class.serializer() as KSerializer<Any>,
+			requestBody
+		)
```
Maxr1998, line 22:
```suggestion
		return json.encodeToString(
			requestBody::class.serializer() as KSerializer<Any>,
			requestBody,
		)
```
or
```suggestion
		return json.encodeToString(requestBody::class.serializer() as KSerializer<Any>, requestBody)
```

jellyfin-api/src/jvmMain/kotlin/org/jellyfin/sdk/api/client/KtorClient.kt
```diff
@@ -121,4 +105,10 @@ public actual open class KtorClient actual constructor(
 			throw err
 		}
 	}
+
+	private fun HttpMethod.asKtorHttpMethod(): io.ktor.http.HttpMethod = when (this) {
```
Maxr1998, line 109: Maybe do an alias-import on `io.ktor.http.HttpMethod` as `KtorHttpMethod`? Because this is rather ugly..

Maxr1998 (changes requested):

Maxr1998:
As always, mostly nit-picking. The rest looks really good, nice work!

Maxr1998 (approved):

PR #324, opened 2021-09-17
Add Api instances to ApiClient as extension properties

nielsvanvelzen:
This change generates extension properties on the ApiClient class to more easily use the API. It comes with some benefits:
- No longer need to remember api names, just use autocomplete/intellisense to find them!
- More memory safe - just a single instance per ApiClient instead of multiple instances created everywhere in the code

The commits are a bit messy so this PR should be squashed when merging.


nielsvanvelzen:
TL;DR code like this:

```kt
val userApi = UserApi(api)
val auth by userApi .authenticateUserByName(...)
```

now becomes this:
```kt
val auth by api.userApi.authenticateUserByName(...)
```

Maxr1998 (approved):

PR #334, opened 2021-11-06
Fix generator ignoring non-json request body types

nielsvanvelzen:
Thanks to @crobibero adding the new logging endpoint which made me check if I needed to change the SDK for the UploadFile operation.

I found a bug where the generator assumed body types to be JSON, and if it was not they would be ignored. supported types are now:

- JSON - like previously
- Images - uses the ByteArray type
- Strings - required for for 10.8 >=

The generator now properly checks the request type and throws if it can't deal with it. This immediately fixes a few operations for image uploads.

Additionally I re-ordered `MimeTypes` alphabetically.


openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
```diff
@@ -63,6 +64,39 @@ class OpenApiApiServicesBuilder(
...
+		val requestBodyTypes = requestBody.content.keys.filterNot {
+			MimeType.IGNORED_JSON_TYPES.contains(it)
+		}
+
+		// Check amount of types
+		if (requestBodyTypes.size > 1) throw OpenApiGeneratorError("Multiple request body types are unsupported.")
+		else if (requestBodyTypes.isEmpty()) return null
+
+		// Retrieve info
+		val requestBodyType = requestBodyTypes.first()
```
Maxr1998, line 78: Could maybe merge these into a neat `when` on `requestBodyTypes.size` for readability, like
```kt
		// Assert count and retrieve type
		val requestBodyType = when (requestBodyTypes.size) {
			0 -> return null
			1 -> requestBodyTypes.first()
			else -> throw OpenApiGeneratorError("Multiple request body types are unsupported.")
		}
```

openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
```diff
@@ -63,6 +64,39 @@ class OpenApiApiServicesBuilder(
...
+		return when (requestBodyType) {
+			// JSON body - use type builder
+			MimeType.APPLICATION_JSON -> content?.schema?.let { schema ->
+				openApiTypeBuilder.build(
+					ApiTypePath(serviceName, operationName, ApiTypePath.PARAMETER_BODY),
+					schema
+				).copy(nullable = !required)
+			}
+			// Image body
+			MimeType.IMAGE_ALL -> Types.BYTE_ARRAY
```
Maxr1998, line 92: This woudln't handle something like `image/png`, would it? Iirc Ktor has a MimeTypes parser/matcher that can also match the 'all' type properly. Not sure whether that could be necessary for us in the future (at least right now it apparently isn't).

openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
```diff
@@ -63,6 +64,39 @@ class OpenApiApiServicesBuilder(
 		else -> null
 	}
 
+	private fun buildRequestBody(requestBody: RequestBody, serviceName: String, operationName: String): TypeName? {
+		// Filter out duplicate JSON values
+		val requestBodyTypes = requestBody.content.keys.filterNot {
+			MimeType.IGNORED_JSON_TYPES.contains(it)
+		}
```
Maxr1998, line 71: Instead of filtering them out, they should maybe be mapped to the `application/json`, and stored in a set for deduplication. Imo that would be a bit cleaner.

openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
nielsvanvelzen: You're correct, the server spec always uses `image/*` right now so I used a strict check.

openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
nielsvanvelzen: All the 3 json types are 100% the same content so we can just remove those other two

Maxr1998: If the main one is always passed, that's ok then.

openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiApiServicesBuilder.kt
Maxr1998: Alright, just something to keep in mind for the future then.

Maxr1998 (approved):

PR #335, opened 2021-11-07
Add extensions to read arguments from GeneralCommandMessage

nielsvanvelzen:
GeneralCommandType is the most often used WebSocket message from the server but it's actual informative content (callled arguments) is just a map<string,string>. There is no schema for it (not even in the server!).

To make it easier to read information I added some (operator) extension functions that are case insensitive to allow reading the command messages.

The tests use a valid DISPLAY_MESSAGE command.

--

This is one of multiple pull requests that will rewrite the WebSocket API to make it easier to work with.

### DIFF
--- a/jellyfin-model/src/commonMain/kotlin/org/jellyfin/sdk/model/extensions/GeneralCommandMessageExtensions.kt
+++ b/jellyfin-model/src/commonMain/kotlin/org/jellyfin/sdk/model/extensions/GeneralCommandMessageExtensions.kt
@@ -1,0 +1,24 @@
+package org.jellyfin.sdk.model.extensions
+
+import org.jellyfin.sdk.model.socket.GeneralCommandMessage
+import kotlin.reflect.KProperty
+
+// Single get
+public operator fun GeneralCommandMessage.get(name: String): String? =
+	arguments.entries.firstOrNull { (key) ->
+		key.contentEquals(name, ignoreCase = true)
+	}?.value
+
+// Multi get
+public operator fun GeneralCommandMessage.get(vararg names: String): List<String?> =
+	names.map(::get)
+
+// Property delegation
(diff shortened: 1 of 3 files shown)


jellyfin-model/src/commonMain/kotlin/org/jellyfin/sdk/model/extensions/GeneralCommandMessageExtensions.kt
```diff
@@ -0,0 +1,18 @@
+package org.jellyfin.sdk.model.extensions
+
+import org.jellyfin.sdk.model.socket.GeneralCommandMessage
+import kotlin.reflect.KProperty
+
+// Single get
+public operator fun GeneralCommandMessage.get(name: String): String? =
+	arguments.entries.firstOrNull { it.key.contentEquals(name, true) }?.value
```
Maxr1998, line 8: Should maybe multi-line this, make the `it` specific, and prepend a `ignoreCase=` before the `true`.

jellyfin-model/src/commonMain/kotlin/org/jellyfin/sdk/model/extensions/GeneralCommandMessageExtensions.kt
```diff
@@ -0,0 +1,18 @@
...
+
+// Multi get
+public operator fun GeneralCommandMessage.get(vararg names: String): List<String?> = names.map(::get)
+
+// Property delegation
+public operator fun GeneralCommandMessage.getValue(thisRef: Any?, property: KProperty<*>): String? = get(property.name)
+
+// Value checking
+public operator fun GeneralCommandMessage.contains(name: String): Boolean =
+	arguments.any { it.key.contentEquals(name, true) }
```
Maxr1998, line 18: Same as above.

Maxr1998 (changes requested):

Maxr1998 (approved):

PR #338, opened 2021-12-09
Add compatibility util to collect flows from Java

GiviMAD:
I was trying to use this sdk on a Java project and I was stuck on how to implement the local server discovery because the api returns a Flow<ServerDiscoveryInfo>.
After some looking on stack overflow I saw the best approach is to add a compatibility util into the project.
So I have added this util, now I can implement the server discovery like this:
```java
         new JavaFlow<ServerDiscoveryInfo>().collect(
                new LocalServerDiscovery(options.build()).discover(30, 5),
                ()->{
                    logger.info("Discovery begins");
                },
                (info) -> {
                    logger.info("Server found: {} - {} - {}", info.getId(), info.getName(), info.getAddress());
                },
                (thr) -> {
                    if (thr != null) {
                        logger.error("Something fail: {}", thr.getMessage());
                        return;
                    }
                    logger.info("Discovery ends without errors");
                }
        );
```
I hope it's an acceptable implementation and could be merged if not I can open an issue.
Let me know if any changes are required.
Regards.


nielsvanvelzen (dismissed):
Thanks for this change, I made some small changes to have it look more like the existing JavaCallback for the API.
It works like this now:

```java
        JavaFlow.<ServerDiscoveryInfo>collect(
                jellyfin.getDiscovery().discoverLocalServers(30, 5),
                () -> logger.info("Discovery begins"),
                (server) -> logger.info("Server found: {} - {} - {}", server.getId(), server.getName(), server.getAddress()),
                (error) -> {
                    if (error != null) {
                        logger.error("Something went wrong", error);
                        return;
                    }
                    
                    logger.info("Discovery ends without errors");
                }
        );
```

Instead of constructing a new JavaFlow object you can use the static function `.collect()`. It returns an instance of `FlowJob` that has the `close` function for that particular flow.

The `@JvmOverloads` annotation generates multiple Java functions so you can call it with just a flow+resultcallback or add all other parameters.

Also, in your sample you constructed the LocalServerDiscovery class directly, while this is possible I suggest using it via the Jellyfin instance like my sample above.

nielsvanvelzen (approved):

Maxr1998 (approved):

nielsvanvelzen (approved):

PR #341, opened 2021-12-18
add java utility to consume suspend functions and move package

GiviMAD:
Signed-off-by: Miguel Álvarez Díez <miguelwork92@gmail.com>

Hi again. 
Still missing another util to consume all the sdk utilities from Java. I have added it next to the JavaFlow and I've moved the package with the two utils to the root of the core.
I have tried to implement it similar to what we did with the JavaFlow, it can be used as follows:
```java
suspendFn(..., JavaContinuation.get((err, result)->{
            // ...
}));
```
Hope everything is correct.
Regards

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=9afb3ca688fc12c761a0e9e4321e4d24e977a4a8916c8a768b1fe05ddb4d6b66
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.1-bin.zip
+distributionSha256Sum=23b89f8eac363f5f4b8336e0530c7295c55b728a9caa5268fdd4a532610d5392
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.2-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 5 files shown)


nielsvanvelzen:
We already have the `JavaCallback` for all API operations. What places are you trying to use that need another continuation implementation?

GiviMAD:
I can't use the recommendation functions for the discovery nether the publish function in the WebsocketApi. Also with JavaCallback and JavaDataCallback I can not access the exceptions in case them occur, which is desirable some places of my code.

GiviMAD:
I have made some changes to make it more similar to the JavaCallback util.
If you think is better that I close the pr and wait for a different solution let me know.

nielsvanvelzen:
Thanks for the explanation. I'd like to have consistency between our various continuation implementations. After your explanation I thought of a few things I want to change:

- Add a way to receive errors in all continuations (JavaCallback/JavaDataCallback)
  - Optional to implement, defaults to throwing
- Add a new continuation for non-api functions (this PR)

For the second item I'm thinking of changing JavaCallback to work for all types, not only Response<...> types.

It will look something like this:

```java
// Based on the java-cli sample code

userApi.authenticateUserByName(new AuthenticateUserByName("demo", null, ""), new JavaResponseCallback<AuthenticationResult>() {
    @Override
    public void onResponse(@Nullable AuthenticationResult response) {
        System.out.println("Access token is " + response.getAccessToken());
    }

    @Override
    public void onError(@Nullable Throwable error) {
        System.err.println("Something went wrong!!");
        error.printStackTrace();
    }
});
```

I'll thinker a bit about it and create a new pull request with those changes. Let me know if you have any thoughts about it.

GiviMAD:
The idea sounds great, so I'll close this one. Thanks for the help.

PR #344, opened 2021-12-23
Disable wildcard imports

nielsvanvelzen:
I remember not doing this before because the editorconfig didn't work in IntelliJ/Android Studio. It works now though. I think it's because of the `ij_kotlin_packages_to_use_import_on_demand` rule.

- Enable Detekt `WildcardImport` rule
- Update `.editorconfig` to "disable" automatic wildcard imports in IntelliJ based editors
  - Uses a value of 1000 imports from the same package, which should hopefully never be reached in a single file
  - Set the packages to always use wildcard imports to an empty string
- Updated all project imports to remove wildcards


Maxr1998 (approved):
That's pretty nice, I didn't know about the `ij_kotlin_packages_to_use_import_on_demand` rule. Have to use that in my own projects :eyes:

nielsvanvelzen:
Yeah it took some time to figure those out, there's not much documentation about Intellij's rules unfortunately.

PR #346, opened 2021-12-25
Rewrite Java callback helpers to support error handling and non-api functions

nielsvanvelzen:
This PR improves the compatibility with Java programs.

- All Java compatibility helpers are now jvmMain / jvmCommonMain
- JavaCallback renamed to JavaResponseContinuation
- JavaDataCallback renamed to JavaContinuation
  - Also works for all suspending functions now, instead of just `Response<T>`
- Added new `onError` method to the continuation helpers
  - Allows for handling errors
  - Optional to implement (although recommended)
  - Throws in the default implementation
- Updated java-cli demo

Additional thanks to @GiviMAD for their help.

Supersedes #341


jellyfin-core/src/jvmCommonMain/kotlin/org/jellyfin/sdk/compatibility/JavaContinuation.kt
```diff
@@ -0,0 +1,16 @@
...
+import kotlin.coroutines.EmptyCoroutineContext
+
+public abstract class JavaContinuation<T> : Continuation<T> {
+	override val context: CoroutineContext = EmptyCoroutineContext
+	override fun resumeWith(result: Result<T>): Unit = result.fold(
+		onSuccess = ::onSuccess,
+		onFailure = ::onError,
+	)
+
+	public abstract fun onSuccess(response: T)
```
Maxr1998, line 14:
```suggestion
	public abstract fun onSuccess(value: T)
```

or "result" or something different.

Maxr1998 (changes requested):
LGTM, the variable could use a different name though.
When you're already at it, I'd put the success case above the error case in the sample.

Maxr1998 (approved):

PR #350, opened 2022-01-05
Support username and password arguments in kotlin-cli sample

nielsvanvelzen:
Wanted to implement this for a while now...

- Support `--username` and `--password` (optional) in kotlin-cli commands as replacement for `--token`
  - When token is not supplied it will sign in to retrieve a token

This is useful for testing various functions of the SDK, like the WebSockets via the observe command using Gradle (`gradlew :samples:kotlin-cli:run --args="observe --server https://demo.jellyfin.org/stable --username demo"`).

### DIFF
--- a/samples/kotlin-cli/src/main/kotlin/org/jellyfin/sample/cli/Options.kt
+++ b/samples/kotlin-cli/src/main/kotlin/org/jellyfin/sample/cli/Options.kt
@@ -3,6 +3,11 @@ package org.jellyfin.sample.cli
 import com.github.ajalt.clikt.core.ParameterHolder
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
+import kotlinx.coroutines.runBlocking
+import org.jellyfin.sdk.Jellyfin
+import org.jellyfin.sdk.api.client.ApiClient
+import org.jellyfin.sdk.api.client.extensions.authenticateUserByName
+import org.jellyfin.sdk.api.client.extensions.userApi
 
 fun ParameterHolder.serverOption() = option(
 	"-s", "--server",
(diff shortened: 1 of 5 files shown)


samples/kotlin-cli/src/main/kotlin/org/jellyfin/sample/cli/Options.kt
```diff
@@ -3,15 +3,50 @@ package org.jellyfin.sample.cli
...
+		help = "Username"
+	).also(::registerOption)
+
+	val password = option(
+		"-p", "--password",
+		help = "Password"
+	).also(::registerOption)
+
+	return lazy {
+		runBlocking {
```
Maxr1998, line 38: Not the biggest fan of the runBlocking, but it's only a sample, I suppose? If you don't have a better alternative on hand already, we can leave it like this.

nielsvanvelzen: All the commands use runBlocking because Clikt is not async

samples/kotlin-cli/src/main/kotlin/org/jellyfin/sample/cli/Options.kt
```diff
@@ -3,15 +3,50 @@ package org.jellyfin.sample.cli
...
+
+	val username = option(
+		"-u", "--username",
+		help = "Username"
+	).also(::registerOption)
+
+	val password = option(
+		"-p", "--password",
+		help = "Password"
+	).also(::registerOption)
```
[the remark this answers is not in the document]
nielsvanvelzen: This function is only called once for each command, just like the options inside. Normally the registerOption() would be called automatically in the back but it didn't in this case, not sure why.

Maxr1998 (approved):

PR #358, opened 2022-03-13
Add MissingBaseUrlException to replace IllegalArgumentException

nielsvanvelzen:
A client often catches the generic ApiClientException when using the API. In some cases the base url might be unset.

This PR adds a new MissingBaseUrlException that is used now. Just like we do for user id's with the MissingUserIdException.


Maxr1998 (approved):

Maxr1998:
Oof, I didn't mean to squash this 😕

PR #373, opened 2022-04-09
Add retry logic to SocketInstance

### LINKED ISSUE #361: Implement reconnect behavior in WebSocket API
Right now when a connection fails the SDK doesn't attempt to reconnect. We should implement some smart reconnect behavior that will attempt to reconnect automatically.

nielsvanvelzen:
Fixes #361 

This PR adds automatic reconnection abilities to the WebSocket code. It's fairly simple for now; retry 5 times when a disconnect/error occurs. Error will wait 30 seconds before retrying while a "normal" disconnect retries after 5. The algorithm could definitely be improved in the future, but  this will do for now.

### DIFF
--- a/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/helper/ReconnectHelper.kt
+++ b/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/helper/ReconnectHelper.kt
@@ -1,9 +1,61 @@
 package org.jellyfin.sdk.api.sockets.helper
 
-internal class ReconnectHelper {
-	fun reportConnect() {}
-	fun reportDisconnect(){}
-	fun reset() {}
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import mu.KotlinLogging
+import kotlin.time.Duration.Companion.seconds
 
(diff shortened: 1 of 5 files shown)


jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/helper/ReconnectHelper.kt
```diff
@@ -1,9 +1,61 @@
...
+			return false
+		}
+
+		reconnectJob = coroutineScope.launch {
+			val retryAfter = if (error) RETRY_ERROR else RETRY_NORMAL
+
+			logger.info { "Reconnect scheduled in $retryAfter (error=$error)" }
+
+			delay(retryAfter)
+			reconnect()
```
Maxr1998, line 50: `reconnect` calls `notificyReconnect` which cancels the Job. Couldn't this cause issues?

nielsvanvelzen: I didn't notice any issues with that while testing

Maxr1998: Alright. Might just be something to be aware of, in case there are any issues in the future.

Maxr1998 (approved):

PR #396, opened 2022-04-30
Migrate unit tests to Kotest

nielsvanvelzen:
- Use kotest framework
- Use kotest assertions
- Add Comparable<ServerVersion> to ServerVersion


Maxr1998 (approved):
Seems good. Didn't do the most accurate review, but it passes, should be ok.

PR #423, opened 2022-06-15
Increase minimum supported Jellyfin version to 10.8.1

nielsvanvelzen:
10.8.1 will add a few changes to the newly introduced Splashscreen API. One of these changes is the possibility to disable the splashscreen completely. This adds a new (not-null) boolean property called `splashscreenEnabled` to the `BrandingOptions` model which is a breaking change. Therefor we'll increase the minimum supported Jellyfin version in a patch update.

The removed import is because the annotation is imported by default and was no longer needed.


nielsvanvelzen:
Note: Do not merge until 10.8.1 is released

nielsvanvelzen:
> Breaking API changes in 10.8.1:
> 
> * `splashscreenEnabled` field in BrandingOptions
> * `vppTonemappingBrightness` and `vppTonemappingContrast` fields in EncodingOptions
> 
> The mentioned API models will cause an exception when using 10.8.0 or lower.

(from #431)

Maxr1998 (approved):

PR #434, opened 2022-07-08
Allow client to change minimum server version

nielsvanvelzen:
This mostly affects the RecommendedServerDiscovery rejecting servers not meeting the hardcoded minimum server version, but the option value can be used by the client too.


Maxr1998 (approved):
LGTM

PR #442, opened 2022-07-19
Map java.net.ConnectException to TimeoutException

AndreasGB:
This exception occurs for me when the server is unreachable, I'm mapping it to `TimeoutException` as it's the general type for network issues.


AndreasGB:
The PR for the Android TV Client that handles the exception in the login flow will probably be submitted this evening

nielsvanvelzen (approved):

Issue #448, opened 2022-08-02
SSL errors not caught


nielsvanvelzen:
Regression from #446, SSL errors are not caught. Ideally we should wrap them in a subclass of ApiClientException. This issue only happens for badly configured servers.

Stacktrace from https://github.com/jellyfin/jellyfin-androidtv/issues/1920.

```
javax.net.ssl.SSLHandshakeException: Handshake failed
	at com.android.org.conscrypt.ConscryptFileDescriptorSocket.startHandshake(ConscryptFileDescriptorSocket.java:286)
	at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.kt:379)
	at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.kt:337)
	at okhttp3.internal.connection.RealConnection.connect(RealConnection.kt:209)
	at okhttp3.internal.connection.ExchangeFinder.findConnection(ExchangeFinder.kt:226)
	at okhttp3.internal.connection.ExchangeFinder.findHealthyConnection(ExchangeFinder.kt:106)
	at okhttp3.internal.connection.ExchangeFinder.find(ExchangeFinder.kt:74)
	at okhttp3.internal.connection.RealCall.initExchange$okhttp(RealCall.kt:255)
	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:32)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:95)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:83)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:76)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:201)
	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:517)
[...]

PR #449, opened 2022-08-02
Add SecureConnectionException for all SSL related exceptions

nielsvanvelzen:
Also catch other IOExceptions

Part 1/2 for #448

### DIFF
--- a/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/exception/SecureConnectionException.kt
+++ b/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/exception/SecureConnectionException.kt
@@ -1,0 +1,10 @@
+package org.jellyfin.sdk.api.client.exception
+
+/**
+ * An error occurred while attempting to establish a secure connection. This can happen when the server doesn't have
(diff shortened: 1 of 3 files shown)


jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/exception/SecureConnectionException.kt
```diff
@@ -0,0 +1,10 @@
+package org.jellyfin.sdk.api.client.exception
+
+/**
+ * An error occurred while attempting a secure connection. This can happen when the server doesn't have HTTPS enabled
```
Maxr1998, line 4:
```suggestion
 * An error occurred while attempting to establish a secure connection. This can happen when the server doesn't have HTTPS enabled
```

Maxr1998 (approved):

Maxr1998 (approved):

nielsvanvelzen:
`(squash)` 👀

Maxr1998:
Do you prefer something else?

nielsvanvelzen:
I don't mind too much, normally I do a rebase for single commits and squash otherwise.

PR #455, opened 2022-08-06
Optimize map building in generator

nielsvanvelzen:
Using buildMap {} with capacity set instead of a mutable map should be more efficient as the capacity is known in advance. I didn't benchmark this though.

### DIFF
--- a/jellyfin-api/src/commonMain/kotlin-generated/org/jellyfin/sdk/api/operations/ActivityLogApi.kt
+++ b/jellyfin-api/src/commonMain/kotlin-generated/org/jellyfin/sdk/api/operations/ActivityLogApi.kt
@@ -9,8 +9,8 @@ import kotlin.Any
 import kotlin.Boolean
 import kotlin.Int
 import kotlin.String
+import kotlin.collections.buildMap
 import kotlin.collections.emptyMap
-import kotlin.collections.mutableMapOf
 import org.jellyfin.sdk.api.client.ApiClient
 import org.jellyfin.sdk.api.client.Response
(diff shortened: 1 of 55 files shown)


Maxr1998:
`mapOf` requires instantiating quite a lot of pairs and an array for them though.

Under the hood, both `mutableMapOf` with no paramters and `mapOf` allocate a LinkedHashMap, passing the count of elements in the second case. By default, those HashMaps allocate more space than for those elements to keep room for future insertions. Thus, in both cases no re-allocation should be necessary, and due to the array creation from the vararg parameter mapOf might even be slower.

Instead, I suggest looking at the documentation of LinkedHashMap, passing parameter to set the exact size needed, and add the elements one after another afterwards.

nielsvanvelzen:
Looks like the Kotlin documentation acknowledges the performance hit of using the `to` infix function and actually recommends the way we did it before: https://kotlinlang.org/docs/constructing-collections.html#construct-from-elements. Now I'm wondering if `buildMap {}` would be more efficient 🤔

Maxr1998:
Looking at [the source](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/collections/builders/MapBuilder.kt), `buildMap` seems to the best option, if you specify the capacity in advance. No overhead from vararg, no pairs, and no overprovisioning of the internal array capacities.

nielsvanvelzen:
Nice timing, I just finished generating the buildMap calls :p

Maxr1998:
I was surprised at how you were already :joy:

Maxr1998 (approved):

Issue #454, opened 2022-08-06
Add data classes for requests with a lot of parameters


nielsvanvelzen:
Some endpoints have a lot of (query-)parameters. Sometimes the values of these parameters change based on some logic. Right now you'll need to write separate function calls or create variables to later pass to the operation function.

A solution for this annoyance would be to add data classes that can be used instead of a big parameter list. They would be an addition to the existing functions, not replace them.

A (simple) example for the "getAudioStreamUrl" operation in the audio api:

```kotlin
// Before
val streamUrl = api.audioApi.getAudioStreamUrl(
	itemId = itemId,
	mediaSourceId = mediaSource.id,
	deviceId = api.deviceInfo.id,
)

// After
var audioStreamRequest = GetAudioStreamRequest(
	itemId = itemId,
	mediaSourceId = mediaSource.id,
	deviceId = api.deviceInfo.id,
)
// Imagine a lot of logic here instead of just this single line
if (someOtherMediaSourceId != null) audioStreamRequest = audioStreamRequest.copy(mediaSourceId = someOtherMediaSourceId)
val streamUrl = api.audioApi.getAudioStreamUrl(audioStreamRequest)
```

The requests should be the exact same as the request parameters but in a data class instead of function parameters.

nielsvanvelzen:
Sample generated function:

```
	public suspend fun getUniversalAudioStream(request: GetUniversalAudioStreamRequest):
			Response<ByteReadChannel> = getUniversalAudioStream(
		itemId = request.itemId,
		container = request.container,
		mediaSourceId = request.mediaSourceId,
		deviceId = request.deviceId,
		userId = request.userId,
		audioCodec = request.audioCodec,
		maxAudioChannels = request.maxAudioChannels,
		transcodingAudioChannels = request.transcodingAudioChannels,
		maxStreamingBitrate = request.maxStreamingBitrate,
		audioBitRate = request.audioBitRate,
		startTimeTicks = request.startTimeTicks,
		transcodingContainer = request.transcodingContainer,
		transcodingProtocol = request.transcodingProtocol,
		maxAudioSampleRate = request.maxAudioSampleRate,
		maxAudioBitDepth = request.maxAudioBitDepth,
		enableRemoteMedia = request.enableRemoteMedia,
		breakOnNonKeyFrames = request.breakOnNonKeyFrames,
		enableRedirection = request.enableRedirection,
	)
```

PR #456, opened 2022-08-07
Support default values in generated models

nielsvanvelzen:
Right now this adds a few booleans/numbers to some response models - nothing fancy. This change is a part of #454 though, where we need the default values generated for request models.

The DefaultValueHook is not supported for models right now.


nielsvanvelzen:
If we add the DefaultValueHook support for models we could technically add some hooks to add 10.8.0 compatibility, although that might be a bit late now..

Maxr1998 (approved):

PR #464, opened 2022-08-15
Don't use Any type for defaultValue in generator

nielsvanvelzen:


Maxr1998 (approved):
LGTM, even though I don't entirely understand all changes, lol.

nielsvanvelzen:
So basically instead of using `Any` we create an sealed interface with the allowed types as subclasses: string, int, number and code block. It makes everything slightly more type safe.

Maxr1998 (approved):
Ok, that makes sense.

PR #465, opened 2022-08-16
Support enum defaults, add generator context

nielsvanvelzen:
- Add generator context
This is a class containing the generated sources in various phases. The context is passed to various builders that need it to either add or read data from it.

- Add DefaultValue.EnumMember
Required for #463. Allows the default value of a property to be an enum member. This needed a type lookup and that's why we now have the context.


nielsvanvelzen:
image

^ With the dependency bump from #463

Maxr1998 (approved):

PR #474, opened 2022-08-25
Remove READ_EXTERNAL_STORAGE permission

### LINKED ISSUE #472: Is permission READ_EXTERNAL_STORAGE required?
First of all thank you so much for making this library, it allowed me to create [Findroid](https://github.com/jarnedemeulemeester/findroid) :smile: 

While looking at the permissions I saw that READ_EXTERNAL_STORAGE is requested.
But is this really required? I removed it from the manifest and everything still seems to work.

According to the docs this is only required when you want to read all contents from external storage and I don't think this library does that. But maybe I'm wrong.
https://developer.android.com/reference/android/Manifest.permission#READ_EXTERNAL_STORAGE

jarnedemeulemeester:
This permission is left-over from the previous apiclient where it was needed to store credential information.
READ_EXTERNAL_STORAGE permission is no longer used so it can be removed.

Closes #472

### DIFF
--- a/jellyfin-core/src/androidMain/AndroidManifest.xml
+++ b/jellyfin-core/src/androidMain/AndroidManifest.xml
@@ -4,5 +4,4 @@
 
 	<uses-permission android:name="android.permission.INTERNET" />
 	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):
Thanks!

PR #475, opened 2022-08-27
Generate request models

### LINKED ISSUE #454: Add data classes for requests with a lot of parameters

nielsvanvelzen:
This change adds a new function to generated api's. The function allows paramaters to be passed to the operation as model (data class) instead of separate parameters. This is especially useful for large queries (retrieving items with certain filters, allowing certain image types, requesting additional fields etc.) that are have a lot of logic to construct.

The functions generate based on hooks. Currently one hook is defined that generates the functions when there are 5 parameters or more. Request models where all values are either nullable or have a default are optional.

Example from the kotlin-cli:

**Before**
```kotlin
val libraries by api.userViewsApi.getUserViews(includeHidden = false)
```

**After**
```kotlin
val librariesRequest = GetUserViewsRequest(includeHidden = false)
val libraries by api.userViewsApi.getUserViews(librariesRequest)
```

Closes #454


Maxr1998 (approved):
LGTM, reviewed the first commit with the generator changes at least.

PR #476, opened 2022-08-27
Build and deploy Dokka documentation in CI

### LINKED ISSUE #211: Publish Dokka documentation
We already add Dokka documentation to the artifacts on Maven Central (docs are required). It would be nice to publish them somewhere more accessible. Something like `developers.jellyfin.org/sdk-kotlin/1.0.0/`.

nielsvanvelzen:
Looking into publishing the markdown docs too, that's a next PR though.

Fixes #211

### DIFF
--- a/.github/workflows/sdk-docs.yaml
+++ b/.github/workflows/sdk-docs.yaml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  build-dokka:
+  build:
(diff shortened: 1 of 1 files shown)


thornbill (commented):
I believe you need a step to configure Pages also https://github.com/jellyfin/jellyfin.org/blob/master/.github/workflows/build.yml#L23

thornbill (approved):

PR #477, opened 2022-08-28
Use gradle/gradle-build-action in CI

nielsvanvelzen:
The gradle-build-action is used to setup CI for Gradle usage and contains better caching capabilities compared to the setup java action.

Also make logs less verbose by removing the --info option


Maxr1998 (approved):
Nice, we should use this for Android too

PR #480, opened 2022-08-28
Build and deploy Vitepress documentation to GitHub pages

nielsvanvelzen:
Also fixes some dead links (that caused CI to fail 😉) and rewrite the existing sdk-docs workflow.

How it works:
- Build Dokka and Vitepress in their own jobs with no references to GH pages. Upload build output as a zip
- Download artifacts from Dokka and Vitepress and put in a directory, upload as GH Pages artifact and deploy

Considerations:
- Vitepress doesn't like being in a directory by default so we need to add a custom domain or change the configuration

### DIFF
--- a/.github/workflows/sdk-docs.yaml
+++ b/.github/workflows/sdk-docs.yaml
@@ -8,6 +8,29 @@ on:
   workflow_dispatch:
 
 jobs:
+  build-vitepress:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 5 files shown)


nielsvanvelzen:
Sample (with broken css etc. because it's a subdir): https://nielsvanvelzen.github.io/jellyfin-sdk-kotlin/ & https://nielsvanvelzen.github.io/jellyfin-sdk-kotlin/dokka/

Maxr1998 (approved):

PR #492, opened 2022-09-06
Improve RecommendedServerDiscovery speed significantly

nielsvanvelzen:
**Improve RecommendedServerDiscovery speed significantly**

Two changes to make it faster:

1. Lower "server not available" timeout from 5 to 3.5 seconds
2. Query 3 candidates simultaneously using a semaphore

Also removed the flow usage and just use a collection now, the flow didn't make much sense anyway. This doesn't break the ATV app and I expect other clients will be fine too.


nielsvanvelzen:
With this change inputting just my local ip address without protocol/port connects to my server in <10 seconds. It took waaaay to long before. This should be a nice quality of life improvement for new users.

thornbill (approved):

PR #519, opened 2022-09-24
Update GitHub actions workflows

nielsvanvelzen:
**Changes**
- Use consistent naming scheme
- Use organization check instead of repository check

**Issues**

### DIFF
--- a/.github/workflows/gradlew-update.yaml
+++ b/.github/workflows/gradlew-update.yaml
@@ -1,4 +1,4 @@
-name: Gradle update
+name: Gradle / Update wrapper
 
 on:
   schedule:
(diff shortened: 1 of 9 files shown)


.github/workflows/gradlew-update.yaml
```diff
@@ -7,11 +7,11 @@ on:
 jobs:
   update:
     runs-on: ubuntu-22.04
-    if: github.repository == 'jellyfin/jellyfin-sdk-kotlin'
+    if: ${{ contains(github.repository_owner, 'jellyfin') }}
```
thornbill, line 10: Why contains instead of equals?

nielsvanvelzen: We match by organization so we can easily copy workflows between repositories without changing too much. This specific line is copied from @h1dden-da3m0n's workflows. I believe we use contains so it still works for the jellyfin-archive and jellyfin-sandbox organizations.

Maxr1998 (approved):

PR #573, opened 2022-10-27
Use multiple query parameters instead of comma separated values in UrlBuilder

nielsvanvelzen:


nielsvanvelzen:
Tested and confirmed working as expected with the Android TV app

Maxr1998 (approved):

PR #649, opened 2023-01-29
refactor: convert Long to Duration for `HttpClientOptions`

### LINKED ISSUE #637: Use Duration instead of Long in HttpClientOptions
We're using Duration in most places so doing it in HttpClientOptions makes sense. Also helps to prevent issues because you can say `30.seconds` instead of `30_000 // seconds`

Yash-Garg:
Closes #637


jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/HttpClientOptions.kt
```diff
@@ -14,17 +17,17 @@ public data class HttpClientOptions(
 	 * Timeout for a complete HTTP request in milliseconds.
 	 * Defaults to 30 seconds.
 	 */
-	val requestTimeout: Long = 30_000,
+	val requestTimeout: Duration = 30.seconds,
```
nielsvanvelzen, line 20: Comments for this and the other timeout variables need to be updated.

nielsvanvelzen (changes requested):

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/HttpClientOptions.kt
Yash-Garg: Done!

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/HttpClientOptions.kt
```diff
@@ -11,20 +14,20 @@ public data class HttpClientOptions(
 	val followRedirects: Boolean = true,
 
 	/**
-	 * Timeout for a complete HTTP request in milliseconds.
+	 * Timeout for a complete HTTP request in seconds.
```
Maxr1998, line 17: Because it's a duration type, there's no unit attached. I think you can simply remove that part (like so):

```suggestion
	 * Timeout for a complete HTTP request.
```

Maxr1998 (changes requested):

PR #684, opened 2023-03-14
Add compare command to openapi-generator

nielsvanvelzen:
I made this months ago but never finished the markdown reporter. So here is everything except the markdown reporter which I'll finish soon ©️.

Right now we're checking the OpenAPI difference in the server repo to help avoid breaking API changes. The tool we use is not perfect and the comments are pretty much unreadable. Thus my idea to do it myself was born. The new subcommand of the openapi-generator is called "compare" and it takes in 2 OpenAPI files and calculated the difference between them. It re-uses our existing code to parse an OpenAPI schema to our intermediate representation and then calculates all differences.

Some use-cases for this tool are:

- View API changes in server pull requests
- Create API changelogs for server releases

- [example JSON report between 10.8.9 and unstable(2023-03-14)](https://gist.github.com/nielsvanvelzen/7940c5a85fa3cf731a2275b641039e45)

### DIFF
--- a/openapi-generator/build.gradle.kts
+++ b/openapi-generator/build.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
 	alias(libs.plugins.download)
+	alias(libs.plugins.kotlin.serialization)
 	kotlin("jvm")
 	id("application")
 }
@@ -22,6 +23,9 @@ dependencies {
 
 	// Kotlin code generation
 	implementation(libs.kotlinpoet)
+
+	// Compare reporters
+	implementation(libs.kotlinx.serialization.json)
 
 	// Dependency Injection
 	implementation(libs.koin)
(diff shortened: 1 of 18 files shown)


thornbill:
Huh why do so many endpoints say they aren't authenticated now?

nielsvanvelzen:
> Huh why do so many endpoints say they aren't authenticated now?

The SDK considers an endpoint as authentication required when they contain the DefaultAuthorization and/or RequiresElevation policy. With https://github.com/jellyfin/jellyfin/pull/9160 the DefaultAuthorization policy is removed from a bunch of endpoints so I need to update the SDK to fix that. Ideally we shouldn't even need to check for RequiresElevation and just consider DefaultAuthorization as auth needed.

PR #694, opened 2023-03-19
add unsupported ssl and server unavailable issues

### LINKED ISSUE #351: Add RecommendedServerIssue for connection problems
Right now connection issues are emitted as MissingSystemInfo, with the throwable pointing to an ApiClientException with a cause containing the actual exception emitted from OkHttp.

**This can be made better/easier.**

I'd suggest creating a new issue type "NoConnection". It should contain the throwable and an enum explaining what went wrong with the connection: bad SSL, no internet etc.

alialiusefi:
Closes #351 
- Add UnsupportedSSL Issue
- Add ServerUnavailable Issue

### DIFF
--- a/jellyfin-api-ktor/src/jvmMain/kotlin/org/jellyfin/sdk/api/ktor/KtorClient.kt
+++ b/jellyfin-api-ktor/src/jvmMain/kotlin/org/jellyfin/sdk/api/ktor/KtorClient.kt
@@ -27,7 +27,7 @@ import org.jellyfin.sdk.api.client.exception.InvalidStatusException
 import org.jellyfin.sdk.api.client.exception.SecureConnectionException
 import org.jellyfin.sdk.api.client.exception.TimeoutException
 import org.jellyfin.sdk.api.client.exception.ssl.BadPeerSSLKeyException
-import org.jellyfin.sdk.api.client.exception.ssl.HandShakeCertificateException
+import org.jellyfin.sdk.api.client.exception.ssl.HandshakeCertificateException
 import org.jellyfin.sdk.api.client.exception.ssl.InvalidSSLProtocolImplementationException
 import org.jellyfin.sdk.api.client.exception.ssl.PeerNotAuthenticatedException
 import org.jellyfin.sdk.api.client.util.ApiSerializer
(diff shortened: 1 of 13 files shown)


jellyfin-core/src/commonMain/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
```diff
@@ -46,10 +48,28 @@ public class RecommendedServerDiscovery constructor(
 		val issues = mutableSetOf<RecommendedServerIssue>()
 		val scores = mutableSetOf<RecommendedServerInfoScore>()
 
+		// Failure checks
+		if (result.systemInfo.isFailure) {
```
nielsvanvelzen, line 52: Should be part of the when statement for system info checks.

nielsvanvelzen (changes requested):
Thanks for the PR. It looks promising so far. I've left two initial about the current approach. Regarding your comment on the issue:

> I'm still thinking on how to approach testing this...

You can likely use https://badssl.com/ to test various SSL issues.

nielsvanvelzen:
@alialiusefi are you still interested in finishing this PR?

alialiusefi:
Hi! yes, i'm still interested. i have been busy with my main commitments sry. I will work on this - this week.

jellyfin-core/src/commonMain/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt
alialiusefi: I looked into the Result class and it's api. Instead of checking if it's null or it's a failure - I just check if there is an exception (implies that there is a failure as per the api).

I also moved the missing system info check to the above let block, otherwise we will have both missing system info issue recommendation and ssl recomendation as well. Here in this case, we are defaulting to MissingSystemInfo issue recomendation when the server replies with something we do not expect (the rest of the known failures line 136) and all other unknown errors.

What do you think?

jellyfin-core/src/commonMain/kotlin/org/jellyfin/sdk/discovery/ssl/SSLIssueResolver.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.jellyfin.sdk.api.client.exception.SecureConnectionException
+import org.jellyfin.sdk.discovery.RecommendedServerIssue
+import javax.net.ssl.SSLHandshakeException
+import javax.net.ssl.SSLKeyException
+import javax.net.ssl.SSLPeerUnverifiedException
+import javax.net.ssl.SSLProtocolException
+
+public class SSLIssueResolver {
+	public companion object {
+		public fun resolveSSLIssue(exception: SecureConnectionException): RecommendedServerIssue {
```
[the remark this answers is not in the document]
alialiusefi: Yup, I agree with that approach.

alialiusefi:
> Thanks for the PR. It looks promising so far. I've left two initial about the current approach. Regarding your comment on the issue:
> 
> > I'm still thinking on how to approach testing this...
> 
> You can likely use https://badssl.com/ to test various SSL issues.

Thanks for the resouce. How realistic do we want to test this and how do we want to do it?

What comes to my mind is to run a badssl docker server when running the test and call the failing endpoints. I'll look into their github page and go with that approach...will need some time to implement the testing part with docker..

alialiusefi:
Hello, I have an update. I was able to use badssl to be able to trigger two custom ssl [exceptions](jellyfin-core/src/jvmTest/kotlin/org/jellyfin/sdk/api/SSLResolverTests.kt). I didnt use a seperate docker badssl container to call as setting it up was time consuming and painful. I suggest to do that in a context of another issue. I instead called badssl.com directly from the test.

jellyfin-api-ktor/src/jvmMain/kotlin/org/jellyfin/sdk/api/ktor/KtorClient.kt
```diff
@@ -136,7 +136,7 @@ public actual open class KtorClient actual constructor(
 			throw InvalidContentException("Serialization failed", err)
 		} catch (err: SSLException) {
 			logger.error(err) { "Unknown SSL error occurred" }
-			throw SecureConnectionException("Unknown SSL error occurred", err)
+			throw SSLExceptionResolver.resolveSSLException(err)
```
nielsvanvelzen, line 139: This can be inlined (e.g. the when cases from SSLExceptionResolver can be added here)

jellyfin-core/src/commonMain/kotlin/org/jellyfin/sdk/discovery/RecommendedServerIssue.kt
```diff
@@ -1,10 +1,20 @@
...
 
+import org.jellyfin.sdk.api.client.exception.SecureConnectionException
 import org.jellyfin.sdk.model.ServerVersion
 
 public sealed interface RecommendedServerIssue {
+
+	/**
+	 * Unsupported SSL. This happens due to incorrect SSL configurations.
+	 */
+	public data class UnsupportedSSL(public val sslException: SecureConnectionException) : RecommendedServerIssue
```
nielsvanvelzen, line 11:
```suggestion
	public data class SecureConnectionFailed(public val exception: SecureConnectionException) : RecommendedServerIssue
```

nielsvanvelzen (changes requested):
Sorry for the delay

alialiusefi:
ok, will look into this

alialiusefi:
Hi! i fixed the comments that you suggested, let me know if i have to update the branch

nielsvanvelzen (approved):
I've made some final adjustments to adhere to our code style and to fix some small issues. Thanks for the pull request!

PR #711, opened 2023-04-08
Add fromNameOrNull and fromName functions to generated enum models

nielsvanvelzen:
The Jellyfin API is a bit messy when working with BaseItemDto. One such case is the "BaseItemDto.status" property which is a different enum based on the item type. For series it refers to `SeriesStatus`. However, we can't use `SeriesStatus.valueOf` because that matches with the SCREAMING_SNAKE_CASE name instead of the PascalCase name.

This PR adds two functions to the companion object of all generated enum members:

- `fromNameOrNull`
  Returns the enum member with the given name using the PascalCase name

- `fromName`
  Calls `fromNameOrNull` and throws when null

~~Additionally this PR fixes a typo in the `MODEL_REQUEST_PARAMETER_DESCRIPTION` string.~~ < moving this one to a separate PR

### DIFF
--- a/openapi-generator/src/main/kotlin/org/jellyfin/openapi/constants/Strings.kt
+++ b/openapi-generator/src/main/kotlin/org/jellyfin/openapi/constants/Strings.kt
@@ -66,4 +66,9 @@ object Strings {
 	 * Description of the parameter used for the generated request model containing operation parameters.
 	 */
 	const val MODEL_REQUEST_PARAMETER_DESCRIPTION = "The request paramaters"
+
+	/**
+	 * Message used in the exception thrown when the serialName is invalid. Use $serialName for the input value.
+	 */
+	const val MODEL_ENUM_MEMBER_EXCEPTION_MESSAGE = "Unknown value \$serialName"
 }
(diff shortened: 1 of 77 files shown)


openapi-generator/src/main/kotlin/org/jellyfin/openapi/constants/Strings.kt
```diff
@@ -66,4 +66,9 @@ object Strings {
 	 * Description of the parameter used for the generated request model containing operation parameters.
 	 */
 	const val MODEL_REQUEST_PARAMETER_DESCRIPTION = "The request paramaters"
+
+	/**
+	 * Message used in the exception thrown when the serialName is invalid. Use $serialName for the input value.
+	 */
+	const val MODEL_ENUM_MEMBER_EXCEPTION_MESSAGE = "Unknown value ${'$'}serialName"
```
Maxr1998, line 73:
```suggestion
	const val MODEL_ENUM_MEMBER_EXCEPTION_MESSAGE = "Unknown value \$serialName"
```

Maxr1998 (changes requested):

PR #745, opened 2023-05-14
Add interfaces and polymorphic serialization to OpenAPI generator

nielsvanvelzen:
- Add support for schema's using `oneOf` without properties
  - Will be created as an (sealed) interface, properties shared between all referenced schemas are added to the interface
- Add support for the `discriminator.propertyName` for interfaces
  - Will enable polymorphic serialization for that interface

This is needed to fully support jellyfin/jellyfin#9682


openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/model/EmptyModelBuilder.kt
```diff
@@ -20,6 +21,15 @@ class EmptyModelBuilder(
 	override fun build(data: EmptyApiModel): JellyFile {
 		return TypeSpec.classBuilder(data.name.toPascalCase(from = CaseFormat.CAPITALIZED_CAMEL))
 			.apply {
+				data.interfaces.forEach { interfaceName ->
+					addSuperinterface(
+						ClassName(
+							Packages.MODEL,
+							interfaceName.toPascalCase(from = CaseFormat.CAPITALIZED_CAMEL)
```
Maxr1998, line 28: It seems that capitalized camel case and pascal case are absolutely identical, so I think this can be dropped.
See: https://github.com/pearxteam/kasechange/blob/6c274238ddae339b7cd0d50751855b710facf223/src/commonMain/kotlin/net/pearx/kasechange/StringExtensions.kt#L40

openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/openapi/OpenApiModelBuilder.kt
```diff
@@ -9,49 +9,111 @@ import org.jellyfin.openapi.model.ApiModel
...
-						deprecated = property.deprecated == true
+	override fun build(context: GeneratorContext, data: Schema<Any>): ApiModel {
+		return when {
+			// Object
+			data.type == "object" -> when {
+				// When properties set, use the object model
+				!data.properties.isNullOrEmpty() -> buildObjectModel(context, data)
+				// When oneOf set, use the interface model
+				!data.oneOf.isNullOrEmpty() -> buildInterfaceModel(context, data)
+				// Otherwise use the object model
```
Maxr1998, line 29:
```suggestion
				// Otherwise use the empty model
```

openapi-generator/src/main/kotlin/org/jellyfin/openapi/model/ObjectApiModel.kt
```diff
@@ -4,5 +4,6 @@ data class ObjectApiModel(
 	override val name: String,
 	override val description: String?,
 	override val deprecated: Boolean,
+	override val interfaces: Set<String>,
 	val properties: Set<ObjectApiModelProperty>
```
Maxr1998, line 8:
```suggestion
	val properties: Set<ObjectApiModelProperty>,
```

PR #746, opened 2023-05-14
Support readOnly properties in OpenAPI generator

nielsvanvelzen:
- Add support for schema properties with readOnly set to true and a default value
  - They will be generated as class property so the serializer will add them and the deserializer ignores them
  - Value can be read by SDK users
  - Will be used for polymorphic discriminator values

This is needed to fully support jellyfin/jellyfin#9682. Sample output: https://github.com/nielsvanvelzen/jellyfin-sdk-kotlin/commit/73ae6251e6cb304f76bee1ebf2992b698a09ab80

Diff for KeepAliveMessage:
```diff
@@ -14,9 +14,10 @@ public data class KeepAliveMessage(
         */
        @SerialName("Data")
        public val `data`: Int,
+) : InboundWebSocketMessage, OutboundWebSocketMessage {
        /**
         * The different kinds of messages that are used in the WebSocket api.
         */
        @SerialName("MessageType")
-       public override val messageType: SessionMessageType = SessionMessageType.KEEP_ALIVE,
-) : InboundWebSocketMessage, OutboundWebSocketMessage
+       public override val messageType: SessionMessageType = SessionMessageType.KEEP_ALIVE
+}
```

PR #772, opened 2023-06-27
Merge EmptyApiModel into ObjectApiModel and add fix empty constructor

nielsvanvelzen:
In #746 support was added for readonly properties. This introduced a previously unnoticed bug where a model with only readonly properties could generate a data class with empty constructor, which is not possible and introduced a compiler error.

To fix this I made 2 changes:
- Merge EmptyApiModel into ObjectApiModel
  This lessens the maintenance burden as both models (and the builders for them) were almost identical
- Add a check to the ObjectModelBuilder to decide whether to include a primary constructor and data class modifier or not.
  This adds the support for empty classes and classes with only readonly properties.

Example of a badly generated class:

```kt
public data class SessionsStopMessage() : InboundWebSocketMessage {
	/**
	 * The different kinds of messages that are used in the WebSocket api.
	 */
	@SerialName("MessageType")
	override val messageType: SessionMessageType = SessionMessageType.SESSIONS_STOP
}
```

And with this PR:

```kt
public class SessionsStopMessage : InboundWebSocketMessage {
	/**
	 * The different kinds of messages that are used in the WebSocket api.
	 */
	@SerialName("MessageType")
	override val messageType: SessionMessageType = SessionMessageType.SESSIONS_STOP
}
```


openapi-generator/src/main/kotlin/org/jellyfin/openapi/builder/model/ObjectModelBuilder.kt
```diff
@@ -135,7 +135,11 @@ class ObjectModelBuilder(
 					)
 				}
 
-				modifiers += KModifier.DATA
+				if (constructor.parameters.isNotEmpty()) {
+					modifiers += KModifier.DATA
+					primaryConstructor(constructor)
+				}
```
nielsvanvelzen, line 141: ^ this is the magic

Maxr1998 (approved):

PR #774, opened 2023-06-28
Update WebSocket code to generated models

nielsvanvelzen:
Thanks to the amazing work from @crobibero we can now (with 10.9) use the WebSocket models from the OpenAPI specification. This PR updates our implementation to use those models.

- Delete manually added WebSocket models
- Update WebSocket code to use generated models
- Catch deserialization issues so the connection doesn't stop forwarding messages

Note: subscriptions still need to be manually added to `subscriptionTypes.kt`, I don't think we can really automate that.

Note2: New models are incompatible with old models because we used different names for some properties. Running the 10.9 spec with a 10.8 server and the other way around mostly works though.

Note3/reminder: **Docs still need updating** <-- 2024 update: will update documentation later ©️ 

~~_Draft until 10.9 beta's start rolling out._~~


nielsvanvelzen:
Made this a year ago.. Didn't think 10.9 would be this far away at that time.

I didn't test it now but as it is mostly switching from the old manually written models to the auto-generated once it should just work. I have some more changes I'd like to make to the WebSocket implementation anyway which I will do in another PR later.

Maxr1998 (approved):
LGTM

PR #796, opened 2023-08-27
Add ignorePathParameters option to UrlBuilder

nielsvanvelzen:
Sometimes you have a string that you don't fully trust and want to build a URL with it. To prevent this string from causing issues there needs to be a way to disable path templates, that way no matter what characters it contains it shouldn't throw an error while keeping the other benefits of the URL builder like properly adding the base URL, fixing double slashes and adding query parameters.

This change is especially useful when working with the `transcodingUrl` returned by the server to construct a URL for the transcoded media.

API changed in a backwards-compatible way.

### DIFF
--- a/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/util/UrlBuilder.kt
+++ b/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/util/UrlBuilder.kt
@@ -19,13 +19,14 @@ public object UrlBuilder {
 		pathTemplate: String = "/",
 		pathParameters: Map<String, Any?> = mapOf(),
 		queryParameters: Map<String, Any?> = mapOf(),
+		ignorePathParameters: Boolean = false,
 	): String {
 		return URLBuilder(baseUrl).apply {
 			// Create from base URL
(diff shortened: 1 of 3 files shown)


jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/util/UrlBuilder.kt
```diff
@@ -64,7 +69,11 @@ public object UrlBuilder {
 
 					lastEnd = i
 				}
+
```
Maxr1998, line 72: If these aren't intended, you could set `ij_kotlin_line_break_after_multiline_when_entry = false` in the `.editorconfig` if you like.

Maxr1998 (approved):

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/util/UrlBuilder.kt
nielsvanvelzen: I actually prefer those so they will slowly get into the codebase :p

PR #841, opened 2023-12-02
Add RegexValidation to OpenAPI generator

nielsvanvelzen:
Jellyfin 10.8.13 introduces some security fixes that restrict the various codec parameters used in the API. This was done using regular expressions that are also added to the OpenAPI specification.

This pull request uses those regex patterns to also validate them client-side to get faster feedback with invalid requests. Just like the previously added IntRangeValidation.

### DIFF
--- a/jellyfin-api/src/commonMain/kotlin-generated/org/jellyfin/sdk/api/operations/AudioApi.kt
+++ b/jellyfin-api/src/commonMain/kotlin-generated/org/jellyfin/sdk/api/operations/AudioApi.kt
@@ -15,6 +15,8 @@ import kotlin.String
 import kotlin.collections.Map
 import kotlin.collections.buildMap
 import kotlin.collections.emptyMap
+import kotlin.require
+import kotlin.text.Regex
 import org.jellyfin.sdk.api.client.ApiClient
 import org.jellyfin.sdk.api.client.Response
 import org.jellyfin.sdk.api.client.extensions.`get`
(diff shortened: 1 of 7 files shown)


jellyfin-api/src/commonMain/kotlin-generated/org/jellyfin/sdk/api/operations/AudioApi.kt
```diff
@@ -150,6 +152,11 @@ public class AudioApi(
 		val pathParameters = buildMap<String, Any?>(1) {
 			put("itemId", itemId)
 		}
+		require(container == null || Regex("""^[a-zA-Z0-9\-\._,|]{0,40}$""").matches(container)) { """Parameter "container" must match ^[a-zA-Z0-9\-\._,|]{0,40}$.""" }
```
Maxr1998, line 155: I know it's because of the generator, but I'm really not a fan of this repeated Regex object creation.
Maybe you could bundle an object class in the SDK which a) validates a certain string against a given string pattern and b) keeps a map of patterns and their respective Regex instances to use. You could also inline the `require` into that new helper function.

Maxr1998: My idea was to just use the pattern as a key. That should work and result in very few RegEx instances.

nielsvanvelzen: So something like this?

```kotlin
object RegexPatterns {
    val `^[a-zA-Z0-9\-\._,|]{0,40}$` = Regex("""^[a-zA-Z0-9\-\._,|]{0,40}$""")
}
```

Maxr1998: No, sorry, I wasn't obvious in my explanation. A code example should do:
```kt
object RegexValidator {
    private val regexPatterns = HashMap<String, Regex>()

    fun validate(pattern: String, input: String): Boolean {
        val regex = regexPatterns.getOrPut(pattern) { Regex(pattern) }
        return regex.matches(input)
    }
}
```

Maxr1998: Fine for now, but raised an issue to improve this later: #864.

Maxr1998 (approved):

PR #870, opened 2024-02-06
Ignore failure of apiDump task in OpenAPI update workflows

nielsvanvelzen:
Currently the unstable branch is not running because it is failing, so I disabled it over 8 months ago. This is not ideal as I normally use the diff to keep an eye on API changes, I prefer reading them in Kotlin instead of JSON.

This PR changes the workflows to ignore the failure of the `apiDump` task, committing code that will fail to build and when pushed, fails various CI tasks. It does allow me to use the unstable branch for diffing though.


nielsvanvelzen:
I'll keep the unstable workflow disabled and manually trigger it when I need it to prevent notification spam for failing CI.

PR #871, opened 2024-02-07
Remove support for includeCredentials from ApiClient.createUrl and generated API

nielsvanvelzen:
We've had support for generating URL's for resources for quite some time now. It supported an option to automatically add the access token to the URL.
This was nice to have but is extremely unsafe, especially when the default value is `true` (which it is for endpoints that require auth). Most often you will generate a URL because you need to pass it to some external library/application. Adding an access token there is a security nightmare. Additionally, URL's often end up in logs.

To encourage safer use of the API (e.g. authorize via headers) we'll remove the option of adding credentials automatically. From my initial checking we never used includeCredentials in the Android TV app (haven't checked mobile but I assume it will be the same). In cases where the access token is still required in the URL the developer will need to add it manually, for that we'll keep the `QUERY_ACCESS_TOKEN` constant in `ApiClient`.

~~Keeping as draft while I think some more about this change, one thing I'm debating is adding an "additionalQueryParameters" option to the generated API functions to make it slightly easier to add the API token manually.~~


nielsvanvelzen:
I'm comitted to this change now. I don't think we need to add the "additionalQueryParameters" argument either. One thing we will need to fix is the mobile app using the `includeCredentials` in a single place: https://github.com/jellyfin/jellyfin-android/blob/b9e154b09442d9263a30cb3855abe0ee72ee4c33/app/src/main/java/org/jellyfin/mobile/player/audio/car/LibraryBrowser.kt#L405

It should be fairly easy to fix this by using authorization via headers in ExoPlayer.

Maxr1998 (approved):

PR #874, opened 2024-02-14
Remove measureBitrate and detectBitrate extensions from MediaInfoApi

nielsvanvelzen:
The implementations of these functions are not reliable. It caused quite a lot of problems in our official Android TV app previously, it was used to determine the maximum bitrate for video playback but ended up setting it too low a lot of the times. This already led us to change to a static value by default a few releases back.


nielsvanvelzen:
Whoops... forgot these were used in the cli samples. Will update the PR later.

Maxr1998 (approved):

PR #879, opened 2024-02-27
Remove DefaultUserIdHook

nielsvanvelzen:
Starting with Jellyfin 10.9 our API is getting a bit smarter and now scores an IQ of about 70 using the SB5 classification. The user id is no longer required for almost all API operations, this includes the query parameters, request bodies and routes (we've added new routes to replace the old ones). Because of this we can finally drop a hack the SDK uses to hide the fact our API is _great_: the DefaultUserIdHook, which automagically added the userId to requests.

First commit: generator changes
Second commit: update generated sources
Third commit: update kotlin-cli sample


nielsvanvelzen:
Marking it as ready now since 1.5.0 will use the 10.9 API specification.

Maxr1998 (approved):

PR #911, opened 2024-04-27
Replace property setters in ApiClient with new update function

nielsvanvelzen:
I've been working on a new WebSocket client implementation for some weeks now and updating credentials for the socket has always been annoying (you need to manually call an `updateCredentials()` function).

This pull request adds the initial changes required to fix this problem. The update function will make the websocket  (if active) reconnect in my wip branch. The property setters are deprecated and should be removed in 1.6.

The reason for adding an update function instead of doing it from the setters it that in most cases you will be updating multiple properties at the same time, which would cause the WebSocket connection to initiate a reconnect multiple times if no extra code was added to prevent that.

Depends on #910

### DIFF
--- a/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/ApiClient.kt
+++ b/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/ApiClient.kt
@@ -8,6 +8,10 @@ import org.jellyfin.sdk.model.ClientInfo
 import org.jellyfin.sdk.model.DeviceInfo
 import kotlin.reflect.KClass
 
+private const val UPDATE_DEPRECATION_MESSAGE =
+	"This property should not be set directly as changes are not propagated to the WebSocket connection." +
+		" Migrate to using the update() function instead."
+
 public abstract class ApiClient {
 	public companion object {
 		/**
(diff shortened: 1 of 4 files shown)


jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/ApiClient.kt
```diff
@@ -8,6 +8,10 @@ import org.jellyfin.sdk.model.ClientInfo
 import org.jellyfin.sdk.model.DeviceInfo
 import kotlin.reflect.KClass
 
+private const val UPDATE_DEPRECATION_MESSAGE =
+	"This property should not be set directly as it does not propagate to all usages." +
+		" Migrate to use the update() function instead."
```
Maxr1998, line 13:
```suggestion
		" Migrate to using the update() function instead."
```

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/ApiClient.kt
```diff
@@ -8,6 +8,10 @@ import org.jellyfin.sdk.model.ClientInfo
 import org.jellyfin.sdk.model.DeviceInfo
 import kotlin.reflect.KClass
 
+private const val UPDATE_DEPRECATION_MESSAGE =
+	"This property should not be set directly as it does not propagate to all usages." +
```
Maxr1998, line 12: usages? Doesn't seem that clear to me yet.
The idea is that the setters don't properly react to changes (e.g. reconnect 
the WebSocket later on), while the update function does, right?

Maxr1998 (changes requested):
Code looks good, deprecation message doesn't seem that clear to me yet.

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/ApiClient.kt
nielsvanvelzen: Yes, although we might use it in other places eventually so I intentionally kept it a bit vague.

Maxr1998: Makes sense. The "propagate to all usages" (of what) could still be improved, maybe :sweat_smile:

Maxr1998 (approved):

PR #913, opened 2024-04-30
Rewrite WebSocket code to be simpler and coroutine based

### LINKED ISSUE #890: SocketInstance / ReconnectHelper can get stuck disconnected
I've been investigating an [issue](https://github.com/jellyfin/jellyfin-androidtv/issues/3461) in jellyfin-androidtv where the websocket disconnects and can't be reconnected.  It appears to be related to the `attempts` in ReconnectHelper.

Our theory is that networking gets suspended once an androidtv app is in the background for a while (seems to be 15-20 min).  When this happens I can see in the logs that the websocket is trying to reconect but it can't.  After 16 reconnect attemps it just stops and there's no way to clear that `attempts` variable externally without making a new SocketInstance.  The only way `attempts` gets reset are the following:

* The socket reconnects
* There are no more listeners attached to the socket and `SocketInstance.updateConnectionState()` gets called

It would be helpful if we could reset that `attempts` counter and reconnect when the jellyfin-androidtv resumes.

Any guidance here?  Or I can just make a PR if this is the proper approach

nielsvanvelzen:
Had some plans to overhaul the WebSocket implementation for a while now and finally started on it a few weeks back. This has been tested with the ATV client (in my 10.9 update branch) and it's working great so far.

The main changes are:
- Event listeners replaced with flows
- Automatically goes idle when there is no active subscriptions
- Automatically goes idle when there is no server/access token set
- Automatically reconnects when server/access token changes (via ApiClient.update function)
- Single instance per ApiClient
- New option to set the reconnect policy/behavior
- Made "publish" function private as there is no use-case for a client to publish
- Greatly improved KDoc documentation
- Optimized performance

Fixes #890.


jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/SubscriptionExtensions.kt
```diff
@@ -0,0 +1,67 @@
...
+public inline fun <reified T : OutboundWebSocketMessage> SocketApi.subscribe(): Flow<T> = subscribe(T::class)
+
+/**
+ * Subscribe to specific [GeneralCommandType] messages.
+ */
+@Suppress("NOTHING_TO_INLINE")
+public inline fun SocketApi.subscribeGeneralCommands(
+	commands: Set<GeneralCommandType> = GeneralCommandType.values().toSet(),
+): Flow<GeneralCommandMessage> = subscribe<GeneralCommandMessage>()
+	.filter { message -> message.data?.name in commands }
```
Maxr1998, line 25: I personally prefer this style in this case, but both would work.
```suggestion
): Flow<GeneralCommandMessage> = subscribe<GeneralCommandMessage>().filter { message ->
	message.data?.name in commands
}
```

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/SocketConnectionState.kt
```diff
@@ -0,0 +1,25 @@
...
+
+	/**
+	 * A connection is currently in progress.
+	 */
+	public data object CONNECTING : SocketConnectionState()
+
+	/**
+	 * The last received message in an active connection.
+	 */
+	public data class MESSAGE(val message: String) : SocketConnectionState()
```
Maxr1998, line 24: These classes should probably use CamelCase.

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/SocketApiState.kt
```diff
@@ -0,0 +1,21 @@
...
+
+	/**
+	 * A connection is currently in progress.
+	 */
+	public data object CONNECTING : SocketApiState()
+
+	/**
+	 * There is an active connection.
+	 */
+	public data object CONNECTED : SocketApiState()
```
Maxr1998, line 20: These classes should probably use CamelCase.

Maxr1998 (changes requested):
Functionality-wise, everything seems good to me :+1:
The comments are mostly style-related. I'm not the biggest fan of using underscores as a variable prefix outside of [backing properties for public fields](https://kotlinlang.org/docs/coding-conventions.html#names-for-backing-properties).

Appreciate the tests for the reconnect policy, btw, and having multiple reconnect policies in the first place. Really elegant design!

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/SubscriptionExtensions.kt
nielsvanvelzen: I find each "action" on its own line more readable

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/DefaultSocketApi.kt
```diff
@@ -0,0 +1,338 @@
...
+
+	/**
+	 * Supervisor scope used for coroutines launched as part of the WebSocket connection.
+	 */
+	private val _scope = CoroutineScope(Dispatchers.IO + SupervisorJob())
+
+	/**
+	 * Instance of the actual socket connection.
+	 */
+	private val _socketConnection = socketConnectionFactory.create(api.httpClientOptions, _scope)
```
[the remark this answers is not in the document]
Maxr1998: Fair enough. Just feel like it would improve readability a bit, not just here, but also for `state` and `messages`.

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/SubscriptionExtensions.kt
Maxr1998: Alright :+1:

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/SocketApiState.kt
Maxr1998: Still applies :eyes:

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/DefaultSocketApi.kt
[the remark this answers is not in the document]
nielsvanvelzen: I did add it for messages now, as that one is quite long so seeing the output type does help. I still think it's not needed for socketConnection though.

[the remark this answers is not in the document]
Maxr1998: Fair, that's where I liked to have it the most. For socketConnection, it's not strictly necessary.

Maxr1998 (approved):

PR #929, opened 2024-05-18
Follow HTTP redirect in remote discovery

### LINKED ISSUE #471: Follow index redirect in discovery
Sometimes people use a baseurl and only have a redirect on the index page (e.g. jellyfin.local -> jellyfin.local/jellyfin). When the user inputs jellyfin.local we should send a HEAD request and see if there is any redirect. If the redirect is the same origin but with a subpath we should add that to the candidates.

This should allow easier setup in Jellyfin apps. Behavior can be tested with our own demo server which redirects to the /stable path.

AllegraCodes:
Optionally add the result of an HTTP redirect to the discovered servers during the remote discovery process, but only if the redirect stays on the same domain.

Resolves #471 

Currently done by adding a HEAD request method to the KtorClient, because the generic request method requires an OK status. If this will remain, there is the opportunity to extract common code to reduce the current duplication, but this may also change as a result of bringing this code in line with the existing api patterns.


AllegraCodes:
If I'm understanding the api code generation correctly, Api objects like systemApi that do most/all of the requests currently are generated based off the server api, so we can't add a head request to those, and therefore it's ok to do the head request directly through the KtorClient

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/ApiClient.kt
```diff
@@ -91,8 +91,16 @@ public abstract class ApiClient {
 		pathParameters: Map<String, Any?> = emptyMap(),
 		queryParameters: Map<String, Any?> = emptyMap(),
 		requestBody: Any? = null,
+		expectedResponseCodes: IntRange = 200 until 300
 	): RawResponse
 
+	public abstract suspend fun headRequest(
```
nielsvanvelzen, line 97: The ApiClient should only have one request function, no additional one should be added just for HEAD requests.

AllegraCodes: On review, I think this PR is a little heavy handed in general. I'm going to start over now that I understand the code better and use a lighter touch.

AllegraCodes:
see #943

PR #942, opened 2024-05-27
Add homedia as third party client

valmnt:
- Update `getting-started.md` to add Homedia as third party client


nielsvanvelzen (approved):

nielsvanvelzen (approved):
Awesome! I believe this is the first third-party Android TV app for Jellyfin.

PR #946, opened 2024-05-31
Support media type for image uploads

### LINKED ISSUE #945: Using client.imageApi.postUserImage reports 400 error
Failed to upload the image using binary. Changed contentType to image/png, converted the binary into Base64 string and passed it into body. Then the upload was successful.
I tested the code that can be uploaded successfully:
```Kotlin
val avatar: String = user.avatar
val bytes: ByteArray = ImageUtils.readFile(avatar)
val mapOf = mutableMapOf<String, String>()
mapOf["Accept"] = "application/json, application/octet-stream;q=0.9, */*;q=0.8"
mapOf["Authorization"] = AuthorizationHeaderBuilder.buildHeader(
    clientName = apiClient.clientInfo.name,
    clientVersion = apiClient.clientInfo.version,
    deviceId = apiClient.deviceInfo.id,
    deviceName = apiClient.deviceInfo.name,
    accessToken = apiClient.accessToken
)
val execute =
    HttpRequest.post("http://localhost:8008/UserImage?userId=ed39b202-5a0b-43de-934f-c7dd421d1af8")
        .timeout(10000).headerMap(mapOf, false)
        .contentType("image/png; charset=UTF-8")
        .body(Base64.getEncoder().encodeToString(bytes))
        .execute()
val body =
    execute.body();

```

sdk version: 
 jellyfin-core-jvm:1.5.0-beta.1
jellyfin version:
 10.9.2
error code:
```Kotlin
val avatar: String = user.avatar
val bytes: ByteArray = ImageUtils.readFile(avatar)
apiClient.imageApi.postUserImage(
    userId = UUID.fromString(user.userId),
    data =
    bytes
)
```
error:
org.jellyfin.sdk.api.client.exception.InvalidStatusException: Invalid HTTP status in response: 400

nielsvanvelzen:
Starting with Jellyfin 10.9 the server does some additional validation on image uploads. It now requires a `Content-Type` header in the request that contains the media type for one of the supported image formats.

The SDK never bothered with that and always told the server it was uploading binary data (`application/octet-stream`). While this is technically correct, the server does not like it anymore. Causing the API to reject all image uploads made using the SDK.

To fix this issue I've added a new model type called `FileInfo`. This model contains the bytes and an media type. The media type defaults to `application/octet-stream`, which obviously does not work for images.

To upload an image you'll need to convert the ByteArray to a FileInfo object. There are two ways to do this:

1. Construct it manually: `val fileInfo = FileInfo(content, "image/png")`
2. Convert it with an extension function: `val fileInfo = content.toFileInfo("image/png")`

An example of uploading a profile image:

```kotlin
val data = ByteArray() // Somehow loaded from file system or whatever

// Old
api.imageApi.postUserImage(data = data)

// New
api.imageApi.postUserImage(data = data.toFileInfo("image/png"))
```

If you don't know the media type of the image you're uploading or just don't care, no worries. The server never validates the actual file you upload, so you can just always set it to `image/png`. Although I don't recommend this as it may eventually be improved for security purposes.

Fixes #945

### DIFF
--- a/jellyfin-model/src/commonMain/kotlin/org/jellyfin/sdk/model/FileInfo.kt
+++ b/jellyfin-model/src/commonMain/kotlin/org/jellyfin/sdk/model/FileInfo.kt
@@ -1,0 +1,23 @@
+package org.jellyfin.sdk.model
+
+/**
+ * Information about a file used to for uploading.
+ */
+public class FileInfo(
+	/**
+	 * The binary content of the file.
+	 */
+	public val content: ByteArray,
+
+	/**
+	 * The media type of the file. Defaults to "
+	 */
+	public val mediaType: String = "application/octet-stream",
+)
+
+/**
+ * Create a [FileInfo] from this byte array.
+ */
+public fun ByteArray.toFileInfo(
+	mediaType: String = "application/octet-stream",
+): FileInfo = FileInfo(this, mediaType)
--- a/jellyfin-api-ktor/src/jvmMain/kotlin/org/jellyfin/sdk/api/ktor/KtorClient.kt
+++ b/jellyfin-api-ktor/src/jvmMain/kotlin/org/jellyfin/sdk/api/ktor/KtorClient.kt
@@ -34,6 +34,7 @@ import org.jellyfin.sdk.api.sockets.SocketApi
 import org.jellyfin.sdk.api.sockets.SocketConnectionFactory
 import org.jellyfin.sdk.model.ClientInfo
 import org.jellyfin.sdk.model.DeviceInfo
+import org.jellyfin.sdk.model.FileInfo
 import java.io.IOException
 import java.net.ConnectException
 import java.net.UnknownHostException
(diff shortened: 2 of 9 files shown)


jellyfin-model/src/commonMain/kotlin/org/jellyfin/sdk/model/FileInfo.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Information about a file used to for uploading.
+ */
+public class FileInfo(
+	/**
+	 * The binary content of the file.
+	 */
+	public val content: ByteArray,
+
+	/**
+	 * The media type of the file. Defaults to "
```
Maxr1998, line 13: Seems to be missing the default in docs.

Maxr1998 (commented):
Sorry for the late review :sweat:
Nonetheless decided to give it a quick look and found one oversight.

PR #960, opened 2024-06-18
Add LiveTVModelsHook

nielsvanvelzen:
I found some issues with the OpenAPI specification while migrating the Android TV app to the SDK. More specifically, the Live TV timer types have properties that should be nullable but are not.

The Live TV API uses the timer DTO's for creating, updating and reading. In the old java apiclient we only required the program id. Now the channelId, keepUntil, startDate, endDate, priority and status properties are also expected. My best guess is that swashbuckle is making some mistake because all of those types are not primitives (GUID/DateTime/Enum) so it probably skips checking for nullability?

Either way, this is a quickfix so the SDK is correct. We should fix this properly in 10.10.


nielsvanvelzen:
@crobibero you may want to look into this one on the server-side.

thornbill (approved):

PR #969, opened 2024-07-23
Fix LocalDateTime in URLBuilder missing timezone

nielsvanvelzen:
We already have a DateTimeSerializer for the JSON bodies, but when sending a datetime via query parameter we should also make sure to include the timezone.


crobibero (approved):

jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/util/UrlBuilder.kt
```diff
@@ -34,6 +37,7 @@ public object UrlBuilder {
 			// Append query parameters
 			queryParameters
 				.filterNot { it.value == null }
+				.entries
```
nielsvanvelzen, line 40:
```suggestion
```

crobibero (approved):

PR #1013, opened 2024-10-27
Check if Android device name is not blank

jarnedemeulemeester:
We should also check if the string returned from `Settings.Global.DEVICE_NAME` is not blank.
The Jellyfin server requires a non-empty device name: 
```
System.ArgumentException: The value cannot be an empty string. (Parameter 'request.DeviceName')
```

This seems like a rare issue but I have at least one report in Findroid.

Downstream issue: https://github.com/jarnedemeulemeester/findroid/issues/846


nielsvanvelzen (approved):
Thanks for the fix! It will be included in the next release (1.6.1) which will most likely release slightly after Jellyfin 10.10.1 for the recommended server version bump.

PR #1050, opened 2025-03-25
Use ByteArray instead of ByteReadChannel in API responses

nielsvanvelzen:
- Use ByteArray instead of ByteReadChannel in RawResponse
- Move non-json decoding from ApiSerializer to RawResponse
- Update ApiSerializer.decodeResponseBody to receive string input
- Update api generator to use ByteArray instead of ByteReadChannel for binary responses
- Remove ktor dependency on jellyfin:api module

(ByteReadChannel is a Ktor type)

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -45,7 +45,6 @@ kotlinx-coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-t
 kotlinx-serialization-core = { module = "org.jetbrains.kotlinx:kotlinx-serialization-core", version.ref = "kotlinx-serialization" }
 kotlinx-serialization-json = { module = "org.jetbrains.kotlinx:kotlinx-serialization-json", version.ref = "kotlinx-serialization" }
 ktor-core = { module = "io.ktor:ktor-client-core", version.ref = "ktor" }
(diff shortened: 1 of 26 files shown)


crobibero (approved):

Maxr1998:
Doesn't this result in more memory usage, since the whole response is stored in memory now? Especially for large files, this might cause issues..

nielsvanvelzen:
Previously the ByteReadChannel would be fully read to memory when reading the response content (via ApiSerializer) already. So it should be the same as before, just not lazy for the few cases the response is not used immediately by the library user.

Unfortunately I couldn't think of a better way to do this. The ByteReadChannel is part of Ktor, which I wanted to fully remove from all modules (except for the ktor apiclient). The Kotlin multiplatform stdlib does not have a different type I could use. There is kotlinx.io but that's not stable yet.

Maxr1998:
Fair, in cases where it is immediately processed afterwards, there should be no huge difference. In the future, being able to stream data would still be nice — maybe it would make sense to create a ticket for that? Hopefully, `kotlinx.io` will be stable soon.

nielsvanvelzen:
In my current testing on the ATV app with the OkHttp implementation networking became more efficient

Maxr1998:
That's good to hear. I hope I'll find some time to review your OkHttp PR soon.

PR #1058, opened 2025-03-29
Update JVM toolchain to 21

BotBlake:
At the moment the build chain uses an ancient version of the JVM Toolchain.
That causes issues for people with JDK versions any higher then 8.

This likely will cause issues with the automated build process on github.

@nielsvanvelzen 🫶 🗿

### DIFF
--- a/jellyfin-api-ktor/build.gradle.kts
+++ b/jellyfin-api-ktor/build.gradle.kts
@@ -8,7 +8,7 @@ kotlin {
 
 	jvm()
 
-	jvmToolchain(23)
+	jvmToolchain(21)
 
 	sourceSets {
 		all {
--- a/jellyfin-api-ktor/build.gradle.kts
+++ b/jellyfin-api-ktor/build.gradle.kts
@@ -8,7 +8,7 @@ kotlin {
 
 	jvm()
 
-	jvmToolchain(8)
+	jvmToolchain(23)
 
 	sourceSets {
 		all {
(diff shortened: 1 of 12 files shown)


anultravioletaurora:
Niels is the G.O.A.T. :muscle:

BotBlake:
Over 50% success rate in the checks... Looks reasonable 🗿

nielsvanvelzen:
I believe we set the toolchain version to Java 8 for compatibility with Android. A better way to do this would be to bump the toolchain (like you did) and set the jvmTarget to Java 8. That will most likely fix the CI issue as well.

BotBlake:
@nielsvanvelzen, should I downgrade the workflows' Java versions to Java 17 again? I changed all of them to Java 21, except for the one that explicitly checks for Java 17 support.

jellyfin-api-ktor/build.gradle.kts
```diff
@@ -6,9 +8,14 @@ plugins {
 kotlin {
 	explicitApi()
 
+	jvm{
+		compilerOptions{
```
nielsvanvelzen, line 12: Formatting is wrong:

```suggestion
	jvm {
		compilerOptions {
```

jellyfin-api-ktor/build.gradle.kts
```diff
@@ -6,9 +8,13 @@ plugins {
 kotlin {
 	explicitApi()
 
-	jvm()
+	jvm {
+		compilerOptions{
```
nielsvanvelzen, line 12:
```suggestion
		compilerOptions {
```

Still not fixed

nielsvanvelzen:
You forgot to set the jvmTarget in the android compilation target of jellyfin-core.

nielsvanvelzen (approved):

PR #1149, opened 2025-10-02
Check API compatibility using Animal Sniffer

nielsvanvelzen:
Use the [Gradle plugin](https://github.com/xvik/gradle-animalsniffer-plugin) for [Animal Sniffer](https://github.com/mojohaus/animal-sniffer) with [Gummy Bears](https://github.com/open-toast/gummy-bears) to verify the SDK is compatible with Android apps using library desugaring and use API level 19 or newer.


crobibero (approved):
those are some library names

PR #1209, opened 2026-03-30
Populate endpointAddress from UDP packet source in LocalServerDiscovery

baykatre:
I was getting localhost from server discovery on my LAN. Turns out the Jellyfin server doesn't populate `EndpointAddress` in its UDP response — it only sends `Address`, `Id`, and `Name`. Other client SDKs (Python, WebOS) fill `endpointAddress` from the UDP packet source IP on the client side, but the Kotlin SDK doesn't.

This is a one-line change: after deserializing the response, set `endpointAddress` to `packet.address.hostAddress`. The `Address` field is left untouched — consumers can check `endpointAddress` when `Address` contains localhost.

I also opened a server-side fix for the root cause: https://github.com/jellyfin/jellyfin/pull/16515

Addresses https://github.com/jellyfin/jellyfin/issues/15898
Addresses https://github.com/jellyfin/jellyfin/issues/11564

### DIFF
--- a/jellyfin-core/src/jvmCommonMain/kotlin/org/jellyfin/sdk/discovery/LocalServerDiscovery.kt
+++ b/jellyfin-core/src/jvmCommonMain/kotlin/org/jellyfin/sdk/discovery/LocalServerDiscovery.kt
@@ -71,7 +71,12 @@ public actual class LocalServerDiscovery actual constructor(jellyfinOptions: Jel
 			// Read as JSON
 			val info = json.decodeFromString(ServerDiscoveryInfo.serializer(), message)
 
-			info
+			// Populate endpointAddress with the actual source IP of the UDP packet.
+			// The server does not set this field; clients are expected to fill it
+			// from the packet origin so consumers can reach the server by its real
+			// network address instead of whatever the server reports in Address
+			// (which is often localhost or 127.0.0.1).
+			info.copy(endpointAddress = packet.address.hostAddress)
 		} catch (err: SocketTimeoutException) {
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
Thank you for your contribution. Unfortunately we won't be accepting it as it violates our [LLM/"AI" Development Policy](https://jellyfin.org/docs/general/contributing/llm-policies).

--

Open for an actual human to make a similar change.

baykatre:
> Thank you for your contribution. Unfortunately we won't be accepting it as it violates our [LLM/"AI" Development Policy](https://jellyfin.org/docs/general/contributing/llm-policies).
> 
> --
> 
> Open for an actual human to make a similar change.

Thanks for the heads up on the LLM policy @nielsvanvelzen — I've updated the PR description in my own words and fixed the commit author. I understand the code and can explain the change.

I also opened a server-side fix for the same root cause: https://github.com/jellyfin/jellyfin/pull/16515

Would you be open to reopening this, or should I open a new PR?

nielsvanvelzen:
As mentioned, the change could be accepted if a human wrote the code. Claude is not a human.

baykatre:
> As mentioned, the change could be accepted if a human wrote the code. Claude is not a human.

You're right, thanks. The test file slipped past me. I've pushed a cleaner version to the branch just the one-line change. Should update once the PR is reopened.

baykatre:
@nielsvanvelzen I also added a isNullOrBlank() check so it only fills endpointAddress when the server doesn't set it. As far as I can tell the server never populates this field — it's always null in the UDP response. Do you know if endpointAddress is intended to carry the packet source IP, or does it have a different meaning?

PR #1211, opened 2026-03-31
Assign endpointAddress from LocalServerDiscovery

nielsvanvelzen:
The `ServerDiscoveryInfo` model from the server always returns `null` for the `endpointAddress`. Looking through our jellyfin-archive it looks like it is intended to be set by sdks/apiclients/apps when doing discovery, to communicate the address from the client perspective. As the "address" property can be overwritten by the server (e.g. using the published server URL option).

It's up to a client what to do with this value. Our recommendation is to always rely on the `address` field returned by the server first before attempting to use the `endpointAddress`.

Human version of #1209.

### DIFF
--- a/jellyfin-core/src/jvmCommonMain/kotlin/org/jellyfin/sdk/discovery/LocalServerDiscovery.kt
+++ b/jellyfin-core/src/jvmCommonMain/kotlin/org/jellyfin/sdk/discovery/LocalServerDiscovery.kt
@@ -1,5 +1,6 @@
 package org.jellyfin.sdk.discovery
 
+import io.github.oshai.kotlinlogging.KotlinLogging
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.flow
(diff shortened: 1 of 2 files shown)


baykatre:
Taking ownership of the changes you made is wrong. Trying to soften it by calling it the "human version" is also completely inaccurate. I would appreciate it if other reviewers would evaluate it from this perspective. Also, I see you haven't looked at the last revision. And additionally, I asked you a question there to get your opinion. @nielsvanvelzen

nielsvanvelzen:
Your AI generated pull request was interpreted as a bug report/feature request and I've made these changes myself with that in mind. Investigating the purpose of this property, writing the code myself, manually testing it and opening a pull request with a human written description. You do not own any of these changes or the work I did to get to this point.

baykatre:
The whole point of using an LLM is to speed up the process, finding the root cause. That's exactly what I did here. The fact that you arrived at the same solution just proves the original work was solid.

I also revised the PR to comply with the LLM policy after your feedback, rewrote the description in my own words, cleaned up the code and removed the test file. You didn't look at any of that before opening your own version. Closing someone's contribution and then opening the same thing under your name doesn't sit right. @nielsvanvelzen

PR #1249, opened 2026-07-12
Update the unstable openapi workflow to only push updates on changes

nielsvanvelzen:
Mostly copied over from jellyfin/jellyfin-sdk-typescript#1072 but with changes to fit the workflow within this repository.


thornbill (approved):

nielsvanvelzen:
Didn't work because the branch we diff with (upstream unstable) isn't fetched so it can't compare:

> fatal: ambiguous argument 'openapi-unstable': unknown revision or path not in the working tree.

need to fix later :copyright: :tm:

PR #1266, opened 2026-08-08
Fix discovery skipping trailing-slash server URLs

StefanoLazzaroni:
## Summary
- Keep both trailing-slash and non-trailing-slash variants when generating address candidates for non-root paths
- Prefer the form that matches the user input, so reverse proxies that treat `/path` and `/path/` differently can still be discovered
- Leave root URLs normalized without a trailing slash (`https://host/` → `https://host`)

## Motivation
Some reverse-proxy setups (especially Jellyfin under a subpath like `/jellyfin/`) behave differently with and without a trailing slash. Discovery previously always stripped the final `/`, so an address such as `https://example.com/jellyfin/` was only tried as `https://example.com/jellyfin` (plus alternate ports), which could time out or fail.

## Test plan
- [ ] Unit tests in `DiscoveryServiceTests` pass for trailing-slash variants and root URL normalization
- [ ] Manual: enter `https://<host>/<subpath>/` and confirm discovery tries the trailing-slash URL first and can connect
- [ ] Manual: enter `https://<host>/<subpath>` and confirm both variants are still tried, with the non-slash form preferred
- [ ] Manual: enter a root URL like `https://localhost/` and confirm candidates do not keep a trailing slash

### DIFF
--- a/jellyfin-core/src/commonMain/kotlin/org/jellyfin/sdk/discovery/AddressCandidateHelper.kt
+++ b/jellyfin-core/src/commonMain/kotlin/org/jellyfin/sdk/discovery/AddressCandidateHelper.kt
@@ -40,6 +40,10 @@ public expect class AddressCandidateHelper(
 	 * The priority is based on a few rules:
 	 * - HTTPS before HTTP
 	 * - Jellyfin ports before protocol default ports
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
Valid bug (please make a human written report) but we do not accept LLM generated pull requests. Closing as invalid.

PR #1286, opened 2026-09-09
Fix websocket subscription grace period

sorryusernameisalreadytaken:
### Problem

The socket was disconnected immediately when the subscription count reached `0`, even though the shared `messages` flow stays alive for another 5 seconds.

If subscriptions returned during that time, the flow was reused but the socket stayed disconnected.

### Change

Let the existing `WhileSubscribed` grace period handle the socket lifetime instead of disconnecting immediately.

### Testing

Added a regression test for quick re-subscription.

SDK tests, lint, and documentation checks pass.

Also tested the equivalent change with Android TV using 2s, 8s, 30s, and 300s standby periods. All tests passed.

I also verified that Home rows picked up newly added movies and shows again after resume without having to restart the app.

### Related

Addresses the WebSocket lifecycle part of jellyfin/jellyfin-androidtv#1149.

### AI assistance

An LLM was used during analysis and test preparation. I reviewed, understood, and runtime-tested the change myself.

### DIFF
--- a/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/DefaultSocketApi.kt
+++ b/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/sockets/DefaultSocketApi.kt
@@ -220,9 +220,7 @@ public class DefaultSocketApi(
 			_subscriptionCount--
 			logger.info { "Subscription count changed to $_subscriptionCount" }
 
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
While this PR addresses an issue, it didn't fix it completely. I've made a new PR and added you as co-author to the commit for the initial changes. Thanks for working on it!

#1290

PR #1290, opened 2026-09-15
Fix WebSocket subscription cancellation

nielsvanvelzen:
Alternative to #1286 that fixes some additional race conditions during cancellation as well. The stop messages are now always sent when the last subscription is removed, but they are awaited.


anultravioletaurora (approved):
lgtm 🚀

sorryusernameisalreadytaken:
Thx for this pr <3

One thing I noticed during my personal android-tv runtime tests before I did my https://github.com/jellyfin/jellyfin-sdk-kotlin/pull/1286 PR:
_subscriptionCount sometimes dropped below zero (for example -1 to -4) during a single normal suspend/resume cycle. Since the counter and map are still unsyncgronized, do you expect the new cancellation handling to prevent this, or would you consider that a separate race? I am a bit unsure about that.

nielsvanvelzen:
In my testing I did notice there were some race conditions and those were gone after my cancellation changes. But I have not (yet) tested within the Android TV app. All my testing was with the CLI sample in this repo (with changes to test various things ofc).

If you do still see some issues please let me know and I'll see what I can do

